PR #10, opened 2019-11-13
Feature/issue7

iowusu:
I added the options to set Topics as ML collections solving issue 7 https://github.com/BillFarber/kafka-marklogic-connector/issues/7

### DIFF
--- a/config/marklogic-sink-marklogic-ssl.properties
+++ b/config/marklogic-sink-marklogic-ssl.properties
@@ -1,0 +1,78 @@
+# Kafka-specific properties
+
+name=marklogic-sink
+connector.class=com.marklogic.kafka.connect.sink.MarkLogicSinkConnector
+
+# Should only need one task since it's using a WriteBatcher, which is multi-threaded
+tasks.max=1
+
+# Topics to consume from [comma separated list for multiple topics]
+topics=marklogic
+
+
+# MarkLogic connector-specific properties
+
+# A MarkLogic host to connect to. The connector uses the Data Movement SDK, and thus it will connect to each of the
+# hosts in a cluster.
+ml.connection.host=localhost
+
+# The port of a REST API server to connect to.
+ml.connection.port=8000
+
+# Optional - the name of a database to connect to. If your REST API server has a content database matching that of the
+# one that you want to write documents to, you do not need to set this.
+ml.connection.database=Documents
+
+# Optional - set to "gateway" when using a load balancer, else leave blank. See https://docs.marklogic.com/guide/java/data-movement#id_26583 for more information.
+ml.connection.type=
+
+# Either DIGEST, BASIC, CERTIFICATE, KERBEROS, or NONE
+ml.connection.securityContextType=DIGEST
+
+# Set these based on the security context type defined above
+ml.connection.username=admin
+ml.connection.password=admin
+ml.connection.certFile=
+ml.connection.certPassword=
+ml.connection.externalName=
+
+# Set to "true" for a "simple" SSL strategy that uses the JVM's default SslContext and X509TrustManager and a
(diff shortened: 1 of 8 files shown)


iowusu:
This was an actually usecase for my  client where they were using a single connector instance, but sending different data to different topics that the connector consume. They needed to tag the data using the topic name in MarkLogic but ml.document.collection could not satisfy that requirement.

config/marklogic-sink-marklogic-ssl.properties
```diff
@@ -59,6 +58,9 @@ ml.dmsdk.threadCount=8
 # Optional - a comma-separated list of collections that each document should be written to
 ml.document.collections=kafka-data
 
+#Optional - set this to false to prevent marklogic from adding topic as collection on documents
```
rjrudin: I think this is a nice feature to have, but it should default to false. The documentation for all the properties for the connector will list it so that a user is aware of it. 

I'd rename it as well to "ml.document.addTopicToCollections", as "ml.document.collections" is a string value and thus I don't think it should have child properties below it. "addTopicToCollections" can then be the name of the attribute in DefaultSinkRecordConverter, as opposed to "addTopicIndicated", which is not consistent with the property name. 

Also, for consistency with the other comments, there should be a space between "#" and "Optional". And "marklogic" should be "MarkLogic" - although it's not MarkLogic doing the adding, it's the connector. I recommend a comment of:

"# Optional - set this to true so that the name of the topic that the connector reads from is added as a collection to each document inserted by the connector"

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -16,11 +15,16 @@
 public class DefaultSinkRecordConverter implements SinkRecordConverter {
 
 	private DocumentWriteOperationBuilder documentWriteOperationBuilder;
-
 	private Format format;
 	private String mimeType;
-
+	private Boolean addTopicIndicated = true ; 
+	private String  mlcollection;
```
rjrudin: I don't believe this is a safe implementation and thus it needs to be changed. Everything except "mlcollection" is set when the converter is initialized, but then "mlcollection" has the potential to change its value at runtime.

I believe the risk occurs when receiving messages from multiple topics, which the config in marklogic-sink.properties indicates is possible - "topics" has a comment of "Topics to consume from [comma separated list for multiple topics]". 

Thus, there's a race condition here, as the converter could be handling messages from different topics at the same time, and the value of "mlcollection" cannot be guaranteed to be correct for a given sink record.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -31,16 +35,30 @@ public DefaultSinkRecordConverter(Map<String, String> kafkaConfig) {
...
+			return  sinkRecord.topic() + ","+ mlcollection;
+	}
+	else {
+			return mlcollection;
+	}
+}
 	@Override
 	public DocumentWriteOperation convert(SinkRecord sinkRecord) {
-		return documentWriteOperationBuilder.build(toContent(sinkRecord));
+		return documentWriteOperationBuilder
```
rjrudin: I believe this is where you need to add the topic to the list of collections. Instead of modifying the state of DocumentWriteOperationBuilder, which introduces a thread safety issue, I recommend doing the following:

- Call build
- Get the DocumentMetadataWriteHandle from the DocumentWriteOperation
- If it can be cast to a DocumentMetadataHandle, then do so and add the topic to the list of collections
- If it cannot be, log a warning and do nothing

PR #13, opened 2019-12-02
Feature/issue7

iowusu:
Applied Rob recommendations.


src/test/java/com/marklogic/kafka/connect/sink/ConvertSinkRecordTest.java
```diff
@@ -31,12 +31,14 @@ public void allPropertiesSet() {
 		config.put("ml.document.permissions", "manage-user,read,manage-admin,update");
 		config.put("ml.document.uriPrefix", "/example/");
 		config.put("ml.document.uriSuffix", ".json");
+		config.put("addTopicToCollections", "true");
```
aclavio: does this need to be `ml.document.addTopicToCollections`

gradle.properties
```diff
@@ -6,4 +6,4 @@ componentOwner=marklogic
 componentName=kafka-marklogic-connector
 
 # Override this in gradle-local.properties
-kafkaHome=
\ No newline at end of file
+kafkaHome=/Users/iowusu/Dropbox/workFiles/tools/testKafkaHome
```
aclavio: might want to remove this local path

.gitignore
```diff
@@ -4,3 +4,13 @@
 build
 out
 gradle-local.properties
+
+bin/test/logback.xml
```
rjrudin: Just ignore the entire "bin" directory, that'll be helpful for people whose tool defaults to compiling files under "bin" (such as VSCode)

README.md
```diff
@@ -78,6 +78,7 @@ When a document is received and written by the connector, you'll see logging lik
 | ml.dmsdk.batchSize | 100 | Sets the number of documents to be written in a batch to MarkLogic. This may not have any impact depending on the connector receives data from Kafka, as the connector calls flushAsync on the DMSDK WriteBatcher after processing every collection of records. Thus, if the connector never receives at one time more than the value of this property, then the value of this property will have no impact. |
 | ml.dmsdk.threadCount | 8 | Sets the number of threads used by the Data Movement SDK for parallelizing writes to MarkLogic. Similar to the batch size property above, this may never come into play depending on how many records the connector receives at once. |
 | ml.document.collections | kafka-data | Optional - a comma-separated list of collections that each document should be written to |
+|ml.document.addTopicToCollections | False | Optional - set this to true so that the name of the topic that the connector reads from is added as a collection to each document inserted by the connector
```
rjrudin: For consistency, this should be "false", not "False"

config/marklogic-sink-marklogic-ssl.properties
```diff
@@ -59,6 +58,9 @@ ml.dmsdk.threadCount=8
 # Optional - a comma-separated list of collections that each document should be written to
 ml.document.collections=kafka-data
 
+# Optional - set this to true so that the name of the topic that the connector reads from is added as a collection to each document inserted by the connector
+ml.document.addTopicToCollections=true
```
rjrudin: The README says this defaults to "false", but this is set to "true" here - it should be "false" to match the README

src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
```diff
@@ -16,10 +16,18 @@
 
 	private ContentIdExtractor contentIdExtractor = new DefaultContentIdExtractor();
 
-	public DocumentWriteOperation build(AbstractWriteHandle content) {
+	public DocumentWriteOperation build(AbstractWriteHandle content, String topic, Boolean addTopicToCollections ) {
```
rjrudin: This isn't a Kafka-specific class (note the package it's in - I'm hoping to port this to ml-javaclient-util once it matures a bit more in this project), so we don't want Kafka-specific parameters here. 

I think the best approach here is to add a method to DocumentWriteOperationBuilder - build(AbstractWriteHandle, DocumentMetadataHandle). That gives a client a chance to construct a DocumentMetadataHandle and populate it a bit themselves before more permissions/collections/etc are added to it. 

Then build(AbstractWriteHandle) can just call build(AbstractWriteHandle, new DocumentMetadataHandle()). 

That gives you a way to add Kafka-specific collections without adding Kafka-specific code to this class.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -16,11 +15,14 @@
...
 	private DocumentWriteOperationBuilder documentWriteOperationBuilder;
-
 	private Format format;
 	private String mimeType;
-
+	private Boolean addTopicToCollections = true; 
+	
 	public DefaultSinkRecordConverter(Map<String, String> kafkaConfig) {
+		
+		addTopicToCollections = Boolean.parseBoolean((kafkaConfig.get(MarkLogicSinkConfig.DOCUMENT_COLLECTIONS_ADD_TOPIC)).trim());
```
rjrudin: Calling "trim" on this will throw a null-pointer exception if the property is not specified or has no value.

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
```diff
@@ -54,6 +55,7 @@
 		.define(DMSDK_TRANSFORM, Type.STRING, Importance.MEDIUM, "Name of a REST transform to use when writing documents")
 		.define(DMSDK_TRANSFORM_PARAMS, Type.STRING, Importance.MEDIUM, "Delimited set of transform names and values")
 		.define(DMSDK_TRANSFORM_PARAMS_DELIMITER, Type.STRING, ",", Importance.LOW, "Delimiter for transform parameter names and values; defaults to a comma")
+		.define(DOCUMENT_COLLECTIONS_ADD_TOPIC, Type.BOOLEAN, true,Importance.LOW, "Boolean that indicate if topics should be add to documents collections ")
```
rjrudin: The README says the default value is "false", but this is "true" - should have a default value of "false"

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
iowusu: Yeah, but there is always a default value false, even if it is not set. The trim is too help with unplanned spaces around the value .

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
iowusu: resolved

iowusu:
I have resolved all the above concerns.

config/marklogic-sink-marklogic-ssl.properties
```diff
@@ -59,6 +58,9 @@ ml.dmsdk.threadCount=8
 # Optional - a comma-separated list of collections that each document should be written to
 ml.document.collections=kafka-data
 
+# Sets this to true so that the name of the topic that the connector reads from is added as a collection to each document inserted by the connector
+ml.document.addTopicToCollections=false
```
rjrudin: Need to add this to config/marklogic-sink.properties as well, otherwise a null pointer is thrown.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -16,11 +16,14 @@
 public class DefaultSinkRecordConverter implements SinkRecordConverter {
 
 	private DocumentWriteOperationBuilder documentWriteOperationBuilder;
-
 	private Format format;
 	private String mimeType;
-
+	private Boolean addTopicToCollections = true;
```
rjrudin: This should default to false, right?

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
rjrudin: If the property isn't specified - which is currently the case in marklogic-sink.properties - then the following error is thrown:

```
[2019-12-04 10:52:41,600] ERROR WorkerSinkTask{id=marklogic-sink-0} Task threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask:177)
java.lang.NullPointerException
        at com.marklogic.kafka.connect.sink.DefaultSinkRecordConverter.<init>(DefaultSinkRecordConverter.java:25)
```

I recommend for defense purposes, just follow the pattern already in place in that method - i.e.

```
val = kafkaConfig.get(MarkLogicSinkConfig.DOCUMENT_COLLECTIONS_ADD_TOPIC);
if (val != null && val.trim().length() > 0) {
  addTopicToCollections = Boolean.parseBoolean(val.trim());
}
```

I think that's actually more defensive than what's needed - I think you just need "val != null". Either way is fine then.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -31,18 +34,33 @@ public DefaultSinkRecordConverter(Map<String, String> kafkaConfig) {
...
+		return documentWriteOperationBuilder.build(toContent(sinkRecord), addTopicToCollections(sinkRecord.topic(), addTopicToCollections ) );
 	}
 
+	/**
+	 * 
+	 *
+	 * @param topic, addTopicToCollections
+	 * @return  metadata 
+	 */
+	public DocumentMetadataHandle addTopicToCollections (String topic, Boolean addTopicToCollections) {
```
rjrudin: For open source, we have to be careful about what methods are public. As part of semantic versioning, we shouldn't be changing public methods in between minor or bugfix releases. But we can fiddle with protected/private methods without worrying about backwards compatibility. So this should be protected or private - I'd do protected since the other non-public method in this class is protected.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -31,18 +34,33 @@ public DefaultSinkRecordConverter(Map<String, String> kafkaConfig) {
...
+		DocumentMetadataHandle metadata = new DocumentMetadataHandle();
+			if (addTopicToCollections) {
+				metadata.getCollections().addAll(topic);
+			} else {
```
rjrudin: No need for this else method, since nothing's being added. Also, just call "add" above instead of "addAll" - "addAll" implies that "topic" is an array, but it's a single string.

iowusu:
@rjrudin resolved  the last concerns.

BillFarber:
I'm sorry, Isaac. I deleted config/marklogic-sink-marklogic-ssl.properties from the 'develop' branch last night. Now it is causing a merge conflict that I can't resolve. Could you delete config/marklogic-sink-marklogic-ssl.properties and config/marklogic-connect-standalone-kafka-ssl.properties, and update the PR.
Thanks.

iowusu:
@BillFarber  done :)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
rjrudin: Should add ".vscode" to .gitignore so that this isn't version-controlled

config/marklogic-sink.properties
```diff
@@ -54,6 +54,9 @@ ml.dmsdk.threadCount=8
 # Optional - a comma-separated list of collections that each document should be written to
 ml.document.collections=kafka-data
 
+# Sets this to true so that the name of the topic that the connector reads from is added as a collection to each document inserted by the connector
```
rjrudin: Should be "Set", not "Sets"

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
```diff
@@ -54,6 +55,7 @@
 		.define(DMSDK_TRANSFORM, Type.STRING, Importance.MEDIUM, "Name of a REST transform to use when writing documents")
 		.define(DMSDK_TRANSFORM_PARAMS, Type.STRING, Importance.MEDIUM, "Delimited set of transform names and values")
 		.define(DMSDK_TRANSFORM_PARAMS_DELIMITER, Type.STRING, ",", Importance.LOW, "Delimiter for transform parameter names and values; defaults to a comma")
+		.define(DOCUMENT_COLLECTIONS_ADD_TOPIC, Type.BOOLEAN, false,Importance.LOW, "Boolean that indicate if topics should be add to documents collections ")
```
rjrudin: Couple tweaks - could be "Boolean that indicates" (add an "s"), but I'd just do "Indicates if". 

And then the last part needs a little tweaking - I think the full string should be "Indicates if the topic name should be added to the set of collections for a document"

PR #15, opened 2019-12-04
Feature/kst 8 null objects

BillFarber:
Handling null messages gracefully.

### DIFF
--- a/src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
+++ b/src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
@@ -17,6 +17,10 @@ public class DocumentWriteOperationBuilder {
 	private ContentIdExtractor contentIdExtractor = new DefaultContentIdExtractor();
 
 	public DocumentWriteOperation build(AbstractWriteHandle content, DocumentMetadataHandle metadata ) {
+		if (content == null) {
+			return null;
+		}
+
 		if (hasText(collections)) {
 			metadata.getCollections().addAll(collections.trim().split(","));
 		}
--- a/src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
+++ b/src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
@@ -18,7 +18,7 @@ public class DocumentWriteOperationBuilder {
 
 	public DocumentWriteOperation build(AbstractWriteHandle content, DocumentMetadataHandle metadata ) {
 		if (content == null) {
-			return null;
+			throw new NullPointerException("'content' must not be null");
 		}
 
 		if (hasText(collections)) {
--- a/src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
+++ b/src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
@@ -70,7 +70,7 @@ public class DefaultSinkRecordConverter implements SinkRecordConverter {
 	 */
 	protected AbstractWriteHandle toContent(SinkRecord record) {
 		if ((record == null) || (record.value() == null)) {
-			return null;
+			throw new NullPointerException("'record' must not be null, and must have a value.");
 		}
 		Object value = record.value();
 		if (value instanceof byte[]) {
--- a/src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
+++ b/src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
@@ -69,6 +69,9 @@ public class DefaultSinkRecordConverter implements SinkRecordConverter {
 	 * @return
 	 */
 	protected AbstractWriteHandle toContent(SinkRecord record) {
+		if ((record == null) || (record.value() == null)) {
+			return null;
+		}
 		Object value = record.value();
 		if (value instanceof byte[]) {
 			BytesHandle content = new BytesHandle((byte[]) value);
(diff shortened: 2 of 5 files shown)


src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
```diff
@@ -17,6 +17,10 @@
 	private ContentIdExtractor contentIdExtractor = new DefaultContentIdExtractor();
 
 	public DocumentWriteOperation build(AbstractWriteHandle content) {
+		if (content == null) {
```
rjrudin: I actually think a null pointer exception is better here. If null is passed in, it may be masking a bigger problem by returning null in response. I think it's better than it throws a null pointer so that the root problem can be identified - in this case, the root problem is that a sink task can receive a null value, which means we want to handle that in the sink task, not here.

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkTask.java
```diff
@@ -114,7 +114,11 @@ public void put(final Collection<SinkRecord> records) {
 			if (logger.isDebugEnabled()) {
 				logger.debug("Processing record value {} in topic {}", record.value(), record.topic());
 			}
-			writeBatcher.add(sinkRecordConverter.convert(record));
+			if (record.value() != null) {
+				writeBatcher.add(sinkRecordConverter.convert(record));
+			} else {
+				logger.info("Skipping record with null value");
```
rjrudin: I am wondering if "warn" is more appropriate here. But I don't know how or why a sink task would receive null, and how unusual that's considered. If it's a common thing that happens a lot and is just natural, then I'd vote for "debug", since then it's just noise to be ignored. But my hunch is that null being received is not a natural event, and thus I lean towards "warn" so it stands out from info logging.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -50,6 +50,9 @@ public DocumentWriteOperation convert(SinkRecord sinkRecord) {
 	 * @return
 	 */
 	protected AbstractWriteHandle toContent(SinkRecord record) {
+		if ((record == null) || (record.value() == null)) {
```
rjrudin: Same thing here - the root problem is that null can be received by the sink task. And what you have the sink task doing is good - log the condition, and then move on. 

If this class gets a null object, I think a null pointer exception is the way to go. Otherwise, something else will likely blow up later when it tries to handle a "null" AbstractWriteHandle, and it'll be more difficult to debug vs if a null pointer is thrown right away here.

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkTask.java
```diff
@@ -114,7 +114,11 @@ public void put(final Collection<SinkRecord> records) {
 			if (logger.isDebugEnabled()) {
 				logger.debug("Processing record value {} in topic {}", record.value(), record.topic());
 			}
-			writeBatcher.add(sinkRecordConverter.convert(record));
+			if (record.value() != null) {
```
rjrudin: Can the record be null as well? i.e. can the Collection<SinkRecord> contain null objects in it? Seems unlikely, but easy enough to check - and that check should go before the debug logging above.

src/test/java/com/marklogic/kafka/connect/sink/ConvertSinkRecordTest.java
```diff
@@ -77,6 +77,24 @@ public void binaryContent() {
 		assertEquals("hello world".getBytes().length, content.get().length);
 	}
 
+	@Test
+	public void emptyContent() {
```
rjrudin: I recommend moving these tests to be against MarkLogicSinkTask, which is where I think the only null check is needed. Today, we know that if we pass in a SinkRecord with a null value (or a null SinkRecord), we'll get an error. So the test would simulate that and just verify that no error occurs (we can't easily verify that a log message is written).

src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
BillFarber: Done.

src/test/java/com/marklogic/kafka/connect/sink/ConvertSinkRecordTest.java
BillFarber: I have made the recommended changes.

BillFarber:
Rebased from the develop branch after finishing the feature/issue7 merge.

BillFarber:
Rebased from the develop branch after finishing the PR for Upgrading ml-javaclient-util to the latest version.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "interactive"
```
rjrudin: This shouldn't be added - I'm thinking this needs to be rebased against develop, as I think the additions to .gitignore are already there on develop?

BillFarber: I rebased a couple of times (though I'm still not sure I'm doing it correctly).
The changes to .gitignore are on my branch. I think the problem is that .vscode got merged into develop when I accepted Isaac's PR.
I will delete the .vscode directory from both the develop branch and my branch.

BillFarber:
Scraping the previous attempts at rebasing from develop and trying again.

BillFarber:
Ok, I don't know what I did wrong with Isaac's PR, but somehow GitHub merged the wrong commits into develop. I give up until tomorrow.

PR #16, opened 2019-12-05
Adds a jetty servlet producer implementation for use with pushing mes…

aclavio:
Adds a jetty servlet producer implementation for use with pushing messages to Kafka from MarkLogic modules.  Includes a sjs library that can be imported into a project to interface with the producer server.


BillFarber:
I concur with Rob.

PR #24, opened 2020-04-19
Can now run a flow using DHF 5.2.0

BillFarber:
Can now run a flow using DHF 5.2.0


rjrudin:
I think this should be merged into the develop branch first. You only want to merge into the "master" branch when you are publishing a release, which will include bumping up a version number. That way, master always reflects the latest release.

PR #34, opened 2020-07-15
Changes for additional Security options

bgeorge111:


src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
```diff
@@ -17,7 +43,28 @@ public DatabaseClientConfig buildDatabaseClientConfig(Map<String, String> kafkaC
 		DatabaseClientConfig clientConfig = new DatabaseClientConfig();
 		clientConfig.setCertFile(kafkaConfig.get(MarkLogicSinkConfig.CONNECTION_CERT_FILE));
 		clientConfig.setCertPassword(kafkaConfig.get(MarkLogicSinkConfig.CONNECTION_CERT_PASSWORD));
-
+		clientConfig.setTrustManager(new SimpleX509TrustManager());
+		
+		/*
+		 * v1.2.2 changes START
+		 */
+		String ssl = kafkaConfig.get(MarkLogicSinkConfig.SSL);
```
rjrudin, line 51: I recommend moving this into a private method to encapsulate the logic here - e.g. "configureHostnameVerifier". I also would avoid comments like this - if someone wants to know when code was added, they can use their version history tool to figure that out.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
```diff
@@ -1,14 +1,40 @@
 package com.marklogic.kafka.connect;
 
+/*
+ * v1.2.2 changes - START
+ */
```
rjrudin, line 5: As noted below too, I recommend not including these comments. The set of imports will change over time, and the comments will soon not be accurate. And it's not worth spending time trying to maintain these comments - they just don't add any value.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
```diff
@@ -35,10 +82,98 @@ public DatabaseClientConfig buildDatabaseClientConfig(Map<String, String> kafkaC
 
 		String securityContextType = kafkaConfig.get(MarkLogicSinkConfig.CONNECTION_SECURITY_CONTEXT_TYPE).toUpperCase();
 		clientConfig.setSecurityContextType(SecurityContextType.valueOf(securityContextType));
-
+		/*
+		 * v1.2.2 changes START
+		 */		
+		if ("BASIC".equals(securityContextType) ||
```
rjrudin, line 88: Same thing here - toss this into a private method (similar to the existing configureSimpleSsl method - in fact, make it a protected method if that makes it easier to invoke separately as part of a unit test) with a self-documenting name like "configureCustomSslConnection". 

The goal is to avoid buildDatabaseClientConfig becoming hundreds of lines long, which makes it much more difficult to understand.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
```diff
@@ -35,10 +82,98 @@ public DatabaseClientConfig buildDatabaseClientConfig(Map<String, String> kafkaC
...
+								clientKeyStore = KeyStore.getInstance("PKCS12");
+							} catch (KeyStoreException e) {
+								
+								throw new RuntimeException("Unable to get default SSLContext: " + e.getMessage(), e);
+							}
+					        TrustManager[] trust = new TrustManager[] { new SimpleX509TrustManager()};
+							
+					        try (InputStream keystoreInputStream = new FileInputStream(clientConfig.getCertFile())) {
+					            clientKeyStore.load(keystoreInputStream, clientConfig.getCertPassword().toCharArray());
+					        } catch (FileNotFoundException e) {
```
rjrudin, line 105: Since you have the same logic for handling each exception type, just catch Exception and throw a RuntimeException. Also, the message isn't accurate - you're not trying to get the default SSLContext, you're trying to load a configure a custom key store. I would have a general message of "Unable to configure custom SSL connection: " + e.getMessage().

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
```diff
@@ -41,6 +41,13 @@
 	public static final String DOCUMENT_MIMETYPE = "ml.document.mimeType";
 	public static final String DOCUMENT_URI_PREFIX = "ml.document.uriPrefix";
 	public static final String DOCUMENT_URI_SUFFIX = "ml.document.uriSuffix";
+	/*
+	 * v1.2.2 changes
+	 */
+	public static final String SSL = "ml.connection.ssl";
```
rjrudin, line 47: I'd rename this to make it more clear that it's a boolean (in hindsight, I wish "ml.connection.simpleSsl" were "ml.connection.enableSimpleSsl"). I think "ml.connection.enableCustomSsl" would be a good descriptive name.

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
```diff
@@ -41,6 +41,13 @@
...
+	public static final String TLS_VERSION = "ml.connection.tlsVersion";
```
rjrudin, line 48: Because these properties are specific to a custom SSL connection, I'd do "ml.connection.customSsl.tlsVersion". That makes it clear that they're not being used as part of ml.connection.simpleSsl.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
bgeorge111: Will make the changes

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
bgeorge111: Will make the changes

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
bgeorge111: Makes sense. Making the changes.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
bgeorge111: Making the changes

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
bgeorge111: Making the changes

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
bgeorge111: Making the changes

bgeorge111:
> I strongly recommend adding unit tests for the new functionality in the client builder. I think it's worth adding a custom keystore for testing purposes to src/test/resources.

Added new test cases

PR #36, opened 2020-08-17
AVRO, JSON Schema and ProtoBuf support, ID Generation Strategies.

bgeorge111:
JSON Schema and ProtoBuf support, ID Generation Strategies.


gradle.properties
```diff
@@ -1,9 +1,10 @@
...
+group=com.marklogic
+
+version=1.2.3
+
+# For the Confluent Connector Archive
+componentOwner=marklogic
+componentName=kafka-marklogic-connector
+
+# Override this in gradle-local.properties
+kafkaHome=
```
BillFarber, line 10: Please revert the changes to this file. I'll update the version when I build the next version.

src/main/java/com/marklogic/client/ext/document/DefaultContentIdExtractor.java
```diff
@@ -1,13 +1,95 @@
...
+	}
+	
+    private static String bytesToHex(byte[] bytes) {
+    StringBuilder sb = new StringBuilder();
+    for (byte b : bytes) {
+        sb.append(String.format("%02x", b));
+    }
+    return sb.toString();
+	}
+}
```
BillFarber, line 95: Please add a newline to the end of the file

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -73,7 +105,29 @@ protected AbstractWriteHandle toContent(SinkRecord record) {
...
+		if (value instanceof String) {
+			logger.info("String document received. Converting to byte[]");
+			converter = "STRING";
+        }
+		
 		if (value instanceof byte[]) {
+			logger.info("Byte[] received. Proceeding with MarkLogic operations.");
 			BytesHandle content = new BytesHandle((byte[]) value);
 			if (format != null) {
 				content.withFormat(format);
```
BillFarber, line 133: Should this be two new implementations of the SinkRecordConverter class - one for JSON and one for AVRO? Then the correct converter would be specified in the config.

BillFarber:
One more question - Is Protobuf synonymous with Avro? If not, I don't understand it in the PR title.

src/main/java/com/marklogic/client/ext/document/DefaultContentIdExtractor.java
bgeorge111: Done

src/main/java/com/marklogic/client/ext/document/ContentIdExtractor.java
```diff
@@ -1,9 +1,13 @@
...
 
 public interface ContentIdExtractor {
 
 	String extractId(AbstractWriteHandle content);
-	
-}
+	String extractId(AbstractWriteHandle content, String path);
+	String extractId(AbstractWriteHandle content, String[] path);
+	String extractId(AbstractWriteHandle content, String topic, Integer partition, long offset, String idStrategy);
+}
```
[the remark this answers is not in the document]
bgeorge111: Done

src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
```diff
@@ -42,6 +65,39 @@ protected String buildUri(AbstractWriteHandle content) {
...
+	protected String buildUri(AbstractWriteHandle content,String topic, Integer partition, long offset, String idStrategy) throws IOException  {
+		String uri = contentIdExtractor.extractId(content,topic,partition,offset,idStrategy);
+		if (hasText(uriPrefix)) {
+			uri = uriPrefix + uri;
+		}
+		if (hasText(uriSuffix)) {
+			uri += uriSuffix;
+		}
+		return uri;
+	}
```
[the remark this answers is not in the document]
bgeorge111: Makes sense. Done. Pls check.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
```diff
@@ -1,182 +1,181 @@
...
+							}catch (KeyManagementException e) {
+								throw new RuntimeException("Unable to configure custom SSL connection:" + e.getMessage(), e);
+							}
+							clientConfig.setSslContext(sslContext);
+						}
+					} /* End of if ssl */
+			}
+		return clientConfig;
+	}
+}
```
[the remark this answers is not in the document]
bgeorge111: Yes. Some options were not taking from the parameters which is corrected.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
bgeorge111: SinkRecord has a schema part and a value part. For the messages having schema (JSON with schema, AVRO, ProtoBuf), the schema will be populated and value will be of type struct. The fromConnectData method of the kafka Connect converts the value to a String. For JSON (without Schema), value is of type Map. For everything else, the value is of type String or byte[]. 
For basically, the DefaultSinkRecordConverter is handling AVRO, ProtoBuf, JSON with schema, JSON and any other messages into format that is loadable to MarkLogic (ie byte[] with GenericDocumentManager). Schema portion of messages carrying schema are ignored. 
So, I think that the DefaultSinkRecordConverter has a common behavior of converting SinkRecord.value to byte[]. Other implementations in future can be to handle situations where SinkRecord.schema also need to be handled. That was my thinking and the user need not give which Converter to be used because for all messages the same thing is done - Convert SinkRecord.value to byte[]
If message type based implementations are provided, then we should give AVRO, ProtoBuf, JsonWithSchema, JsonWithoutSchema and String. Or a more general classification like MessagesWithSchema or MessagesWithoutSchema.
Let me know. 

ProtoBuf vs AVRO - They are different types of message formats and significantly differ in schema management. 

A protobuf and AVRO schemas for the same JSON is given below 

**JSON**
{"Snack":{"name":"cake","calories":260,"colour":"white", "flavor":"vanilla"}}
**ProtoBuf Schema**
`syntax = "proto3";
message Snack {
message obj {
string name = 1;
double calories = 2;
string colour = 3;
string flavor = 4;
}
obj Snack = 1;
}`

**AVRO Schema**
`{
  "name": "MyClass",
  "type": "record",
  "namespace": "com.acme.avro",
  "fields": [
    {
      "name": "Snack",
      "type": {
        "name": "Snack",
        "type": "record",
        "fields": [
          {
            "name": "name",
            "type": "string"
          },
          {
            "name": "calories",
            "type": "int"
          },
          {
            "name": "colour",
            "type": "string"
          },
          {
            "name": "flavor",
            "type": "string"
          }
        ]
      }
    }
  ]
}`

src/main/java/com/marklogic/client/ext/document/ContentIdExtractor.java
```diff
@@ -1,9 +1,13 @@
...
 import com.marklogic.client.io.marker.AbstractWriteHandle;
 
 public interface ContentIdExtractor {
 
 	String extractId(AbstractWriteHandle content);
-	
-}
+	String extractId(AbstractWriteHandle content, String path);
+	String extractId(AbstractWriteHandle content, String[] path);
+	String extractId(AbstractWriteHandle content, String topic, Integer partition, long offset, String idStrategy);
```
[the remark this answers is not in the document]
bgeorge111: Added. Test  methods in `ConvertSinkRecordTest `for different ID strategies. Pls take a look. The `ContentIdExtractor `is not a functional interface any more, because there are more than one abstract methods. So, the current junit method using lambda had to be refactored. Pls see the changes in `ConvertSinkR.ecordTest.noPropertiesSet` 
Also note that build.gradle should have `testCompile "com.google.code.gson:gson:2.8.6"` because real JSONs should be built for some ID strategies.

gradle.properties
bgeorge111: Can you help on how to revert ?

BillFarber: If you haven't setup a remote that points to the marklogic-community version, setup that remote first:
git remote add upstream https://bgeorge111:<YourGithubPassword>@github.com/marklogic-community/kafka-marklogic-connector.git

Now you can pull the version of the file in that repo:
git fetch --all
git checkout upstream/master -- gradle.properties
git add gradle.properties
git commit -m "reverting gradle.properties"
git push

src/main/java/com/marklogic/client/ext/document/ContentIdExtractor.java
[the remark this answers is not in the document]
BillFarber: Looks good! Thanks.

gradle.properties
bgeorge111: I believe I have done it. https://github.com/marklogic-community/kafka-marklogic-connector/pull/36/commits/b97f9d2e8d4d23c9355b1fa646c6d20c59be9b2d

gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=com.marklogic
-version=1.2.2
+version=1.3.0
```
rjrudin, line 2: I recommend bumping this up in a separate commit. Makes it easy to look at the commit history and see when the version number was changed. 

Or, change this to 1.3.0-SNAPSHOT to capture that this isn't the real 1.3.0 build yet.

src/main/java/com/marklogic/client/ext/document/ContentIdExtractor.java
```diff
@@ -1,9 +1,13 @@
 package com.marklogic.client.ext.document;
 
+import java.io.IOException;
+
 import com.marklogic.client.io.marker.AbstractWriteHandle;
 
+//@FunctionalInterface
```
rjrudin, line 7: Is this needed? If not, I recommend removing this line - commented-out code can cause confusion.

src/main/java/com/marklogic/kafka/connect/sink/SinkRecordConverter.java
```diff
@@ -10,6 +11,6 @@
  */
 public interface SinkRecordConverter {
 
-	DocumentWriteOperation convert(SinkRecord sinkRecord);
+	DocumentWriteOperation convert(SinkRecord sinkRecord)  throws IOException;
```
rjrudin, line 14: What's the benefit of throwing an IOException from this? Can the client possibly do anything with it, other than catching it and logging it? And is there a reason the client should catch and log an IOException but not do that for any other Exception type?

src/main/java/com/marklogic/client/ext/document/DocumentWriteOperationBuilder.java
```diff
@@ -15,32 +17,62 @@
...
 		}
-
-		String uri = buildUri(content);
-		return build(operationType, uri, metadata, content);
-	}
-
-	protected String buildUri(AbstractWriteHandle content) {
-		String uri = contentIdExtractor.extractId(content);
+		String uri = "";
+		if ("JSONPATH".equals(idStrategy)) {
```
rjrudin, line 35: Any time I see a big if/else or switch like this, I think a better approach is to have a separate implementation class for each strategy - https://en.wikipedia.org/wiki/Strategy_pattern - and then have (in this case) a Map<String, ContentIdExtractor> where each key is one of the strings here. 

That forces us to rethink the ContentIdExtractor interface. Instead of adding methods to it, what we want are multiple implementations. 

But, some of the implementations need more information than just the AbstractWriteHandle - they need the topic, partition, and offset. Those are all on the SinkRecord - so add the SinkRecord as an argument to what should be the only method in the ContentIdExtractor interface - e.g. extractId(content, sinkRecord). 

Additionally, the JSONPATH and HASH strategies need one or more paths. Since those don't vary on every call - like the content and sinkRecord do - those should be constructor arguments, not interface method arguments. 

Thus, what DefaultSinkRecordConverter should do is read properties from MarkLogicSinkConfig and use those to build a Map<String, ContentIdExtractor> (which is essentially a lightweight Factory class). That Map is then passed to DocumentWriteOperationBuilder (instead of a single ContentIdExtractor), which will then choose which ContentIdExtractor to use based on the nature of the content... although see below on that, I'm wondering how to avoid those implementation details from leaking out.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -1,20 +1,45 @@
...
+import org.apache.kafka.connect.storage.Converter;
+import java.util.Collections;
+import org.apache.kafka.connect.json.JsonConverter;
 
 /**
  * Handles converting a SinkRecord into a DocumentWriteOperation via the properties in the given config map.
  */
 public class DefaultSinkRecordConverter implements SinkRecordConverter {
 
+	String converter = "STRING";
```
rjrudin, line 35: This doesn't seem safe, as the value of this is being changed. Whereas all the other fields are initialized in the constructor and then remain static. Is MarkLogicSinkTask guaranteed to be single-threaded, such that its "put" method will never be invoked at the same time by two or more threads? Because if the latter happens, this will break. 

Actually, what's this field even used for? I found places where it's set, but I didn't find anywhere that its value is used.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -73,7 +105,29 @@ protected AbstractWriteHandle toContent(SinkRecord record) {
 			throw new NullPointerException("'record' must not be null, and must have a value.");
 		}
 		Object value = record.value();
+		Schema schema = record.valueSchema();
+		/* Determine the converter */
+		if (schema != null && value instanceof Struct) {
```
rjrudin, line 110: I don't understand what's going on here. There doesn't seem to be a need to determine the converter, because "converter" is never used anywhere.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -73,7 +105,29 @@ protected AbstractWriteHandle toContent(SinkRecord record) {
...
+			/* Avro, ProtoBuf or JSON with schema, ignore schema, handle only the value */
+			converter = "AVRO_OR_JSON_WITH_SCHEMA";
+			logger.info("Avro or JsonWithSchema document received. Converting to byte[].");
```
rjrudin, line 113: This seems like debug or more likely trace-level logging. I would either remove it, or if you keep it, provide some context - i.e. some details about the SinkRecord being processed.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
```diff
@@ -73,7 +105,29 @@ protected AbstractWriteHandle toContent(SinkRecord record) {
...
+			final String payload = new String(JSON_CONVERTER.fromConnectData(record.topic(), schema, value), StandardCharsets.UTF_8);
+			value = payload.getBytes();
+		}
+		
+        if (value instanceof Map) {
```
rjrudin, line 118: This is another instance where a Strategy interface could be used, as it's getting a bit more complicated - gotta check for not just a byte[], but a Struct and a Map. 

How much of a connection is there between these strategies and the ones for ContentIdExtractor? I am wondering if we need a separate interface here - e.g. ContentExtractor - which does "AbstractWriteHandle toContent(SinkRecord)" - and the implementations of that may implement ContentIdExtractor as well.

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkTask.java
```diff
@@ -154,7 +155,12 @@ public void put(final Collection<SinkRecord> records) {
 					logger.debug("Processing record value {} in topic {}", record.value(), record.topic());
 				}
 				if (record.value() != null) {
-					writeBatcher.add(sinkRecordConverter.convert(record));
+					try {
+						writeBatcher.add(sinkRecordConverter.convert(record));
+					}
+					catch (IOException e) {
```
rjrudin, line 161: See my comment below - why catch and log but not rethrow only an IOException here?

rjrudin (changes requested):

gradle.properties
BillFarber: "version=1.3.0" makes it match Master. I'll go ahead and bump the version after completing the PR.

src/main/java/com/marklogic/kafka/connect/sink/SinkRecordConverter.java
BillFarber: Clean this up in the follow-up refactor issue.

src/main/java/com/marklogic/kafka/connect/sink/DefaultSinkRecordConverter.java
BillFarber: This will be addressed in the follow-up refactor issue.

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkTask.java
BillFarber: This will be addressed in the follow-up refactor issue.

PR #68, opened 2022-09-01
Added support for Bulk Data Services

### LINKED ISSUE #3: Allow for data to be ingested via a Data Services endpoint
Data Services skips some overhead in the REST processing libraries in both the JVM process and MarkLogic, so will be faster. Since Kafka is Java-based, there's no need for the more generic REST/HTTP communication.

rjrudin:
Resolves #3 . Did a lot of refactoring (described below) to support this, as my initial attempt to cram this into MarkLogicSinkTask resulted in a too-large, too-complicated class.

- Extracted AbstractSinkTask and WriteBatcherSinkTask to capture what MarkLogicSinkTask
- Added BulkDataServicesSinkTask to capture the new functionality
- MarkLogicSinkConnector now looks at the user config to determine what type of task to create. And all of our integration tests now start with this class.
- Moved WriteFailureHandler from its own class into WriteBatcherSinkTask since it is not intended to be used by any other class
- Renamed BuildSuccessListenerTest to BuildRunFlowListenerTest to make it more self-documenting
- Moved a couple of tests from ConvertSinkRecordTest into HandleInvalidSinkRecordTest because they really needed to be integration tests


src/main/java/com/marklogic/kafka/connect/sink/BulkDataServicesSinkTask.java
```diff
@@ -0,0 +1,204 @@
+package com.marklogic.kafka.connect.sink;
```
rjrudin, line 1: This is all the new stuff. I don't think anything else changed in terms of functionality, just had to move a lot of stuff around.

src/main/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConfig.java
```diff
@@ -124,6 +129,11 @@ public class MarkLogicSinkConfig extends AbstractConfig {
         .define(DMSDK_INCLUDE_KAFKA_METADATA, Type.BOOLEAN, null, Importance.LOW,
             "Set to true so that Kafka record metadata is added to document metadata before it is written. If the document fails to be written, the Kafka record metadata will be logged as well.")
 
+        // TODO Need more info here on the API declaration itself?
```
rjrudin, line 132: I am going to resolve this before the 1.7.0 release by having it point to content in the README, just haven't written that content yet.

anu3990 (approved):

PR #87, opened 2022-11-05
KAFKA-86: Bump the dependencies to use the latest version of Kafka

BillFarber:
Updated the project to use Kafka 3.3.1
Added "org.slf4j:slf4j-api:1.7.36" to the dependencies since Kafka Connect no longer includes the library.
Added Jackson 2.13.4 libraries to build.gradle to ensure compatibility with the next version of the MarkLogic Java Client.
Fixed a couple of lint warnings.


build.gradle
```diff
@@ -64,6 +72,11 @@ dependencies {
   assets files('apache_logo.png')
 }
 
+tasks.withType(JavaCompile) {
```
rjrudin, line 75: Add a comment above this explaining why this config is here (I can google for what it does, but why it's being added should be captured here so we remember in 6 months).

PR #88, opened 2022-11-11
DEVEXP-102: Integrate Kafka-Junit into the project for testing

BillFarber:
Integrate Kafka-Junit into the project by creating a simple test of the existing ML connector functionality

### DIFF
--- a/src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorConfigBuilder.java
+++ b/src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorConfigBuilder.java
@@ -1,0 +1,75 @@
+package com.marklogic.kafka.connect.sink;
+
+import org.apache.kafka.connect.runtime.ConnectorConfig;
+
+import java.util.Properties;
+import java.util.UUID;
+
+public class MarkLogicSinkConnectorConfigBuilder {
+
+    private String topic = String.format("topic-%s", UUID.randomUUID().toString());
+
+    private String key = String.format("key-%s", UUID.randomUUID().toString());
+
+    private final Properties connectorProps = new Properties();
+
+    public MarkLogicSinkConnectorConfigBuilder withTopic(final String topic) {
+        this.topic = topic;
+        return this;
+    }
+
+    public MarkLogicSinkConnectorConfigBuilder withKey(final String key) {
+        this.key = key;
+        return this;
+    }
+
+    public <T> MarkLogicSinkConnectorConfigBuilder with(final String propertyName, final T value) {
+        connectorProps.put(propertyName, value);
+        return this;
+    }
+
+    public <T> MarkLogicSinkConnectorConfigBuilder withAll(final Properties connectorProps) {
+        this.connectorProps.putAll(connectorProps);
+        return this;
+    }
+
+    private <T> void ifNonExisting(final String propertyName, final T value) {
+        if (connectorProps.get(propertyName) != null) return;
+        connectorProps.put(propertyName, value);
+    }
+
+    public Properties build() {
+
+        ifNonExisting(ConnectorConfig.NAME_CONFIG, "marklogic-sink");
+        ifNonExisting(ConnectorConfig.CONNECTOR_CLASS_CONFIG, "MarkLogicSinkConnector");
+        ifNonExisting(ConnectorConfig.TASKS_MAX_CONFIG, "1");
+        ifNonExisting("topics", topic);
+        ifNonExisting("key", key);
+
+        ifNonExisting(MarkLogicSinkConfig.CONNECTION_HOST, "localhost");
+        ifNonExisting(MarkLogicSinkConfig.CONNECTION_PORT, "8000");
+        ifNonExisting(MarkLogicSinkConfig.CONNECTION_SECURITY_CONTEXT_TYPE, "DIGEST");
+        ifNonExisting(MarkLogicSinkConfig.CONNECTION_USERNAME, "admin");
+        ifNonExisting(MarkLogicSinkConfig.CONNECTION_PASSWORD, "admin");
+        ifNonExisting(MarkLogicSinkConfig.DOCUMENT_FORMAT, "JSON");
+        ifNonExisting(MarkLogicSinkConfig.DOCUMENT_COLLECTIONS, "kafka-data");
+        ifNonExisting(MarkLogicSinkConfig.DOCUMENT_PERMISSIONS, "rest-reader,read,rest-writer,update");
+        ifNonExisting(MarkLogicSinkConfig.DOCUMENT_URI_PREFIX, "/kafka-data/");
+        ifNonExisting(MarkLogicSinkConfig.DOCUMENT_URI_SUFFIX, ".json");
+        ifNonExisting(MarkLogicSinkConfig.DMSDK_BATCH_SIZE, "100");
+        ifNonExisting(MarkLogicSinkConfig.DMSDK_THREAD_COUNT, "8");
+        ifNonExisting("key.converter", "org.apache.kafka.connect.storage.StringConverter");
+        ifNonExisting("value.converter", "org.apache.kafka.connect.storage.StringConverter");
+        ifNonExisting("errors.log.enable", true);
+
+        final Properties copyOfConnectorProps = new Properties();
+        copyOfConnectorProps.putAll(connectorProps);
+
+        return copyOfConnectorProps;
+    }
+
+    public static MarkLogicSinkConnectorConfigBuilder create() {
+        return new MarkLogicSinkConnectorConfigBuilder();
+    }
+}
+
--- a/src/test/java/com/marklogic/kafka/connect/sink/WriteFromKafkaTest.java
+++ b/src/test/java/com/marklogic/kafka/connect/sink/WriteFromKafkaTest.java
@@ -1,0 +1,84 @@
+package com.marklogic.kafka.connect.sink;
+
+import com.marklogic.client.io.SearchHandle;
+import com.marklogic.client.query.StructuredQueryBuilder;
+import com.marklogic.client.query.StructuredQueryDefinition;
+import kafka.server.KafkaConfig$;
+import net.mguenther.kafka.junit.*;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.*;
+import static net.mguenther.kafka.junit.EmbeddedConnectConfig.kafkaConnect;
+import static net.mguenther.kafka.junit.EmbeddedKafkaCluster.provisionWith;
+import static net.mguenther.kafka.junit.EmbeddedKafkaClusterConfig.newClusterConfig;
+
+public class WriteFromKafkaTest extends AbstractIntegrationTest {
+
+    private final String ML_COLLECTION = "kafka-data";
+    private final String TOPIC = "test-topic";
+    private final String KEY = String.format("key-%s", UUID.randomUUID());
+
+    private EmbeddedKafkaCluster kafka;
+
+    @BeforeEach
+    void setupKafka() {
+        provisionKafkaWithConnectAndMarkLogicConnector();
+        kafka.start();
+    }
+
+    @AfterEach
+    void tearDownKafka() {
+        kafka.stop();
+    }
+
+    @Test
+    void shouldWaitForKeyedRecordsToBePublished() throws InterruptedException {
+        Integer NUM_RECORDS = 2;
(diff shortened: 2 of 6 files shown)


src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorIntegrationTest.java
```diff
@@ -0,0 +1,116 @@
...
+    private final String TOPIC = "test-topic";
+    private final String KEY = String.format("key-%s", UUID.randomUUID());
+
+    private EmbeddedKafkaCluster kafka;
+
+    @BeforeEach
+    void setupKafka() {
+        provisionKafkaWithConnectAndMarkLogicConnector();
+        kafka.start();
+        cleanUpDataForTest();
```
rjrudin, line 40: No need for this, the parent class takes care of this automatically

src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorConfigBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public Properties build() {
+
+        ifNonExisting(ConnectorConfig.NAME_CONFIG, "marklogic-sink");
+        ifNonExisting(ConnectorConfig.CONNECTOR_CLASS_CONFIG, "MarkLogicSinkConnector");
+        ifNonExisting(ConnectorConfig.TASKS_MAX_CONFIG, "1");
+        ifNonExisting("topics", topic);
+        ifNonExisting("key", key);
+
+        ifNonExisting("ml.connection.host", "localhost");
```
rjrudin, line 49: Use the constants in MarkLogicConfig here, just to avoid duplicating the strings (which the tests aren't yet consistent about, but that's a gradual cleanup effort).

src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorConfigBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+        ifNonExisting("ml.connection.username", "admin");
+        ifNonExisting("ml.connection.password", "admin");
+        ifNonExisting("ml.document.format", "JSON");
+        ifNonExisting("ml.document.collections", "kafka-data");
+        ifNonExisting("ml.document.permissions", "rest-reader,read,rest-writer,update");
+        ifNonExisting("ml.document.uriPrefix", "/kafka-data/");
+        ifNonExisting("ml.document.uriSuffix", ".json");
+        ifNonExisting("ml.dmsdk.batchSize", "100");
+        ifNonExisting("ml.dmsdk.threadCount", "8");
+        ifNonExisting("key.converter", "org.apache.kafka.connect.storage.StringConverter");
```
rjrudin, line 61: There are a bunch of warnings like this when the test is run:

    WARN  o.a.k.c.admin.AdminClientConfig - The configuration 'key.converter' was supplied but isn't a known config.

Do you know if those can be ignored? If so, I think it'd be good to add this to logback.xml:

```
<logger name="org.apache.kafka.clients" level="ERROR" additivity="false">
    <appender-ref ref="STDOUT"/>
  </logger>
```

That'll remove all the warnings about unrecognized configs, as the warnings are coming from multiple classes.

src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorIntegrationTest.java
```diff
@@ -0,0 +1,116 @@
...
+    private final String ML_COLLECTION = "kafka-data";
+    private final String TOPIC = "test-topic";
+    private final String KEY = String.format("key-%s", UUID.randomUUID());
+
+    private EmbeddedKafkaCluster kafka;
+
+    @BeforeEach
+    void setupKafka() {
+        provisionKafkaWithConnectAndMarkLogicConnector();
+        kafka.start();
```
rjrudin, line 39: This results in a bunch of stacktraces about optional dependencies not existing. I think we can safely ignore those, and this does the trick in logback.xml:

```
<logger name="org.reflections" level="ERROR" additivity="false">
    <appender-ref ref="STDOUT"/>
  </logger>
```

src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorIntegrationTest.java
```diff
@@ -0,0 +1,116 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+import static net.mguenther.kafka.junit.EmbeddedConnectConfig.kafkaConnect;
+import static net.mguenther.kafka.junit.EmbeddedKafkaCluster.provisionWith;
+import static net.mguenther.kafka.junit.EmbeddedKafkaClusterConfig.newClusterConfig;
+import static net.mguenther.kafka.junit.ObserveKeyValues.on;
+
+public class MarkLogicSinkConnectorIntegrationTest extends AbstractIntegrationTest {
```
rjrudin, line 24: I'm thinking for these tests, we should suffix them with "KafkaTest". And for consistency with other tests (and this is my generally preferred pattern), let's do "(use case under test)KafkaTest" - so e.g. "WriteFromKafkaTest" (the lack of "WriteBatcher" or "BulkDataServices" in the name means we're using WriteBatcher).

src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorIntegrationTest.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @AfterEach
+    void tearDownKafka() {
+        kafka.stop();
+    }
+
+    @Test
+    void shouldWaitForKeyedRecordsToBePublished() throws InterruptedException {
+        Integer NUM_RECORDS = 100;
```
rjrudin, line 50: While the performance of 1 vs 100 docs wasn't very different - like tens to hundreds of ms different, so not much - it's usually much easier to debug something with fewer docs. I don't think we're interested in performance here, but it would be nice to do more than 1 doc, so I think the best number for this test is 2 docs. Fast and easy to debug.

rjrudin (changes requested):
This looks like a useful tool, but I'm a little worried about the performance of the test. The test ran in about 7.7s on my machine, and the bulk of the time was taken up by the following 3 things:

1. About 3.5s to setup
2. About 2s for `assertCanConsumeRecordsFromTheTopic`
3. About 1.7s for `kafka.stop()`

So 5.2s of the 7.7s seems like fixed overhead. 

I'm wondering if `assertCanConsumeRecordsFromTheTopic` matters - it seems that since the test is verifying that there are messages on the topic and that docs end up in ML, that would suffice. Removing that would cut out 2s. Though - I think the existence of that 2s delay is what's ensuring that the "check collection in ML" test is always working, as 2s is plenty of time for 100 docs to be written. 

Can you see if it's possible to do the setup and teardown in static beforeclass/afterclass functions? I think that would require a second simple test to make sure that both tests can pass when the setup/teardown is done at the class level. That would at least avoid the overhead tax on every test method. It would admittedly encourage more of a "kitchen sink" test class, but that might be the right tradeoff to minimize the number of times we wait for setup/teardown of kafka-junit.

BillFarber:
> I'm wondering if `assertCanConsumeRecordsFromTheTopic` matters -
assertCanConsumeRecordsFromTheTopic - It's definitely not needed for this test. I mainly included it because I didn't want to lose the code. I will be useful in my next test, so I will remove it.


> Though - I think the existence of that 2s delay is what's ensuring that the "check collection in ML" test is always working, as 2s is plenty of time for 100 docs to be written.
When I remove assertCanConsumeRecordsFromTheTopic, the call to check the collection in ML is still good. In fact, I added a log statement immediately after sending the messages to the topic. The ML connector finishes processing the messages BEFORE that log statement. So, the connector is delivering the messages as fast as they are added. At least in this test.


> Can you see if it's possible to do the setup and teardown in static beforeclass/afterclass functions?
I shall.

src/test/java/com/marklogic/kafka/connect/sink/MarkLogicSinkConnectorConfigBuilder.java
BillFarber: How about if I do this in logback.xml?
  <root level="WARN">
    <appender-ref ref="STDOUT"/>
  </root>
  <logger name="com.marklogic" level="INFO" additivity="false">
    <appender-ref ref="STDOUT"/>
  </logger>
  <logger name="org.reflections" level="ERROR" additivity="false">
    <appender-ref ref="STDOUT"/>
  </logger>
  <logger name="org.apache.kafka.clients" level="ERROR" additivity="false">
    <appender-ref ref="STDOUT"/>
  </logger>

That gets rid of nearly all the non-marklogic logging, so it's much easier to see what's happening. If a bug requires debugging, the developer can always change the setting to get more information.

rjrudin: That sounds good, and right, developer can always change the log settings temporarily.

src/test/java/com/marklogic/kafka/connect/sink/WriteFromKafkaTest.java
```diff
@@ -0,0 +1,100 @@
...
+import static net.mguenther.kafka.junit.EmbeddedConnectConfig.kafkaConnect;
+import static net.mguenther.kafka.junit.EmbeddedKafkaCluster.provisionWith;
+import static net.mguenther.kafka.junit.EmbeddedKafkaClusterConfig.newClusterConfig;
+import static net.mguenther.kafka.junit.ObserveKeyValues.on;
+
+public class WriteFromKafkaTest extends AbstractIntegrationTest {
+
+    // Declared by AbstractSpringMarkLogicTest
+    @Autowired
+    private SimpleTestConfig testConfig;
```
rjrudin, line 27: Shouldn't need this, can use the one declared in the parent class

src/test/java/com/marklogic/kafka/connect/sink/WriteFromKafkaTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private void assertMarkLogicDocumentsExistInCollection(String collection, Integer numRecords, String message) {
+        StructuredQueryBuilder qb = new StructuredQueryBuilder();
+        StructuredQueryDefinition queryDefinition = qb.collection(collection);
+        SearchHandle results = getDatabaseClient().newQueryManager().search(queryDefinition, new SearchHandle());
+        assertEquals(numRecords.longValue(), results.getTotalResults(), message);
+    }
+
+    private void assertRecordsOnTheTopic(String topic, Integer num_records, Integer timeCt) throws InterruptedException {
```
rjrudin, line 97: Can remove this, as it's unused (though might be worth pasting into the JIRA ticket for future reference if needed in the future).

rjrudin (commented):
I added this to `AbstractIntegrationTest` to try to minimize the sleep overhead and the chance of intermittent failure:

```
private final static long DEFAULT_RETRY_SLEEP_TIME = 250;
private final static int DEFAULT_RETRY_ATTEMPTS = 10;

// And then this all goes at the bottom of the file (remove this comment)
protected final void retryIfNotSuccessful(Runnable r) {
    retryIfNotSuccessful(r, DEFAULT_RETRY_SLEEP_TIME, DEFAULT_RETRY_ATTEMPTS);
}

protected final void retryIfNotSuccessful(Runnable r, long sleepTime, int attempts) {
    for (int i = 1; i <= attempts; i++) {
        logger.info("Trying assertion, attempt " + i + " out of " + attempts);
        try {
            r.run();
            return;
        } catch (Throwable ex) {
            if (i == attempts) {
                throw ex;
            }
            logger.info("Assertion failed: " + ex.getMessage() + "; will sleep for " + sleepTime + " ms and try again");
            try {
                Thread.sleep(sleepTime);
            } catch (InterruptedException e) {
                // Ignore, not expected during a test
            }
        }
    }
}
```

And then you can do this in the test class:

```
    @Test
    void shouldWaitForKeyedRecordsToBePublished() throws InterruptedException {
        Integer NUM_RECORDS = 2;
        sendSomeJsonMessages(NUM_RECORDS);
        retryIfNotSuccessful(() -> assertMarkLogicDocumentsExistInCollection(ML_COLLECTION, NUM_RECORDS,
                format("Expected to find %d records in the ML database", NUM_RECORDS)));
    }
```

Definitely open to a better method name, but figure this is good enough for now. Also, the code I wrote above should probably eventually move to an `AbstractKafkaJunitTest` that extends `AbstractIntegration` but we can cross that bridge later.

rjrudin (approved):

PR #89, opened 2022-11-18
DEVEXP-108: Read rows via Optic DSL statement.

BillFarber:
Initially, just create a barebones Source connector. Querying MarkLogic will come next.

### DIFF
--- a/src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
@@ -1,0 +1,77 @@
+package com.marklogic.kafka.connect.source;
+
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigDef.Importance;
+import org.apache.kafka.common.config.ConfigDef.Type;
+
+import java.util.Map;
+
+/**
+ * Defines configuration properties for the MarkLogic source connector.
+ */
+public class MarkLogicSourceConfig extends AbstractConfig {
+
+    public static final String CONNECTION_HOST = "ml.connection.host";
+    public static final String CONNECTION_PORT = "ml.connection.port";
+    public static final String CONNECTION_DATABASE = "ml.connection.database";
+    public static final String CONNECTION_SECURITY_CONTEXT_TYPE = "ml.connection.securityContextType";
+    public static final String CONNECTION_USERNAME = "ml.connection.username";
+    public static final String CONNECTION_PASSWORD = "ml.connection.password";
+    public static final String CONNECTION_TYPE = "ml.connection.type";
+    public static final String CONNECTION_SIMPLE_SSL = "ml.connection.simpleSsl";
+    public static final String CONNECTION_CERT_FILE = "ml.connection.certFile";
+    public static final String CONNECTION_CERT_PASSWORD = "ml.connection.certPassword";
+    public static final String CONNECTION_EXTERNAL_NAME = "ml.connection.externalName";
+    public static final String ENABLE_CUSTOM_SSL = "ml.connection.enableCustomSsl";
+    public static final String TLS_VERSION = "ml.connection.customSsl.tlsVersion";
+    public static final String SSL_HOST_VERIFIER = "ml.connection.customSsl.hostNameVerifier";
+    public static final String SSL_MUTUAL_AUTH = "ml.connection.customSsl.mutualAuth";
+
+    public static final String DMSDK_BATCH_SIZE = "ml.dmsdk.batchSize";
+    public static final String DMSDK_THREAD_COUNT = "ml.dmsdk.threadCount";
+
+    public static ConfigDef CONFIG_DEF = new ConfigDef()
+        .define(CONNECTION_HOST, Type.STRING, Importance.HIGH,
+            "Required; a MarkLogic host to connect to. By default, the connector uses the Data Movement SDK, and thus it will connect to each of the hosts in a cluster.")
+        .define(CONNECTION_PORT, Type.INT, Importance.HIGH,
+            "Required; the port of a REST API app server to connect to")
+        .define(CONNECTION_SECURITY_CONTEXT_TYPE, Type.STRING, "DIGEST", Importance.HIGH,
+            "Required; the authentication scheme used by the server defined by ml.connection.port; either 'DIGEST', 'BASIC', 'CERTIFICATE', 'KERBEROS', or 'NONE'")
+        .define(CONNECTION_USERNAME, Type.STRING, null, Importance.MEDIUM,
+            "MarkLogic username for 'DIGEST' and 'BASIC' authentication")
+        .define(CONNECTION_PASSWORD, Type.PASSWORD, null, Importance.MEDIUM,
+            "MarkLogic password for 'DIGEST' and 'BASIC' authentication")
+        .define(CONNECTION_CERT_FILE, Type.STRING, null, Importance.MEDIUM,
+            "Path to PKCS12 file for 'CERTIFICATE' authentication")
+        .define(CONNECTION_CERT_PASSWORD, Type.PASSWORD, null, Importance.MEDIUM,
+            "Password for PKCS12 file for 'CERTIFICATE' authentication")
+        .define(CONNECTION_EXTERNAL_NAME, Type.STRING, null, Importance.MEDIUM,
+            "External name for 'KERBEROS' authentication")
+        .define(CONNECTION_DATABASE, Type.STRING, null, Importance.LOW,
+            "Name of a database to connect to. If your REST API server has a content database matching that of the one that you want to read documents from, you do not need to set this.")
+        .define(CONNECTION_TYPE, Type.STRING, null, Importance.MEDIUM,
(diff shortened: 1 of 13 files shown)


src/main/java/com/marklogic/kafka/connect/source/AbstractSourceTask.java
```diff
@@ -0,0 +1,66 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Base class for concrete SourceTask implementations, providing some generic functionality.
+ */
+abstract class AbstractSourceTask extends SourceTask {
```
rjrudin, line 14: Let's not make this until we have a second type of source task (odds are we won't, at least not for awhile).

src/main/java/com/marklogic/kafka/connect/source/AbstractSourceTask.java
```diff
@@ -0,0 +1,66 @@
...
+
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private boolean logKeys = false;
```
rjrudin, line 18: Let's not include these yet - I'm not sure if they'll have any value anyway since we're not going to populate either on the SourceRecords that we make.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.kafka.common.config.ConfigDef.Type;
+
+import java.util.Map;
+
+/**
+ * Defines configuration properties for the MarkLogic source connector.
+ */
+public class MarkLogicSourceConfig extends AbstractConfig {
+
+    public static final String CONNECTION_HOST = "ml.connection.host";
```
rjrudin, line 15: Doesn't need to happen yet, but I think we'll want a base class of `MarkLogicConfig` with the common properties (probably just connection-related stuff) which extends `AbstractConfig`.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -0,0 +1,87 @@
...
+    public static final String CONNECTION_EXTERNAL_NAME = "ml.connection.externalName";
+    public static final String ENABLE_CUSTOM_SSL = "ml.connection.enableCustomSsl";
+    public static final String TLS_VERSION = "ml.connection.customSsl.tlsVersion";
+    public static final String SSL_HOST_VERIFIER = "ml.connection.customSsl.hostNameVerifier";
+    public static final String SSL_MUTUAL_AUTH = "ml.connection.customSsl.mutualAuth";
+
+    public static final String DMSDK_BATCH_SIZE = "ml.dmsdk.batchSize";
+    public static final String DMSDK_THREAD_COUNT = "ml.dmsdk.threadCount";
+
+    public static final String BULK_DS_ENDPOINT_URI = "ml.source.bulkds.endpointUri";
```
rjrudin, line 34: Don't need the bulk stuff (hopefully we never do, as based on the DHF Spark 2 code, it's pretty difficult to custom code this stuff)

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -0,0 +1,87 @@
...
+    public static final String CONNECTION_SIMPLE_SSL = "ml.connection.simpleSsl";
+    public static final String CONNECTION_CERT_FILE = "ml.connection.certFile";
+    public static final String CONNECTION_CERT_PASSWORD = "ml.connection.certPassword";
+    public static final String CONNECTION_EXTERNAL_NAME = "ml.connection.externalName";
+    public static final String ENABLE_CUSTOM_SSL = "ml.connection.enableCustomSsl";
+    public static final String TLS_VERSION = "ml.connection.customSsl.tlsVersion";
+    public static final String SSL_HOST_VERIFIER = "ml.connection.customSsl.hostNameVerifier";
+    public static final String SSL_MUTUAL_AUTH = "ml.connection.customSsl.mutualAuth";
+
+    public static final String DMSDK_BATCH_SIZE = "ml.dmsdk.batchSize";
```
rjrudin, line 31: My thought in the story for this was to do `ml.source.optic.batchSize` and `ml.source.optic.threadCount` so that all properties pertaining to reading rows via optic are prefixed under `ml.source.optic`. We could reuse `ml.dmsdk.batchSize`, since RowBatcher is part of DMSDK, but I was thinking the benefit of all the source-specific properties having the same prefix outweighs the benefit of reusing the ml.dmsdk properties. Thoughts?

src/main/java/com/marklogic/kafka/connect/source/RowRecordConverter.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.marklogic.client.row.RowRecord;
+import org.apache.kafka.connect.source.SourceRecord;
+
+/**
+ * Defines how a MarkLogic RowRecord is converted into a Kafka SourceRecord, which can then be
+ * sent to a Kafka Topic. Simplifies testing of this logic, as this avoids any dependency on a running MarkLogic
+ * or running Kafka instance.
+ */
+public interface RowRecordConverter {
```
rjrudin, line 11: I think let's hold off on this as we're hoping this is a very simple operation - i.e. "new SourceRecord(topic, rowRecordAsJson)". Whereas on the "write" side, there's a lot of complexity to hide with converting a SinkRecord into a DocumentWriteOperation.

src/test/java/com/marklogic/kafka/connect/source/WriteToKafkaTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Properties;
+
+import static net.mguenther.kafka.junit.EmbeddedConnectConfig.kafkaConnect;
+import static net.mguenther.kafka.junit.EmbeddedKafkaCluster.provisionWith;
+import static net.mguenther.kafka.junit.EmbeddedKafkaClusterConfig.newClusterConfig;
+
+public class WriteToKafkaTest extends AbstractIntegrationSourceTest {
```
rjrudin, line 15: Let's use "Read" here so that "Write" exclusively means "Kafka to ML" and "Read" exclusively means "ML to Kafka". Assuming we won't have any "Read" tests that don't involve kafka-junit, I think we can just do e.g. "ReadRowsViaOpticDslTest" and then "ReadRowsViaSerializedPlanTest".

src/test/java/com/marklogic/kafka/connect/source/WriteToKafkaTest.java
```diff
@@ -0,0 +1,68 @@
...
+        );
+    }
+
+    private Properties sourceConnectorConfig(final String topic, final String opticDsl) {
+        return MarkLogicSourceConnectorConfigBuilder.create()
+            .withTopic(topic)
+            .withDsl(opticDsl)
+            .with(MarkLogicSourceConfig.CONNECTION_HOST, testConfig.getHost())
+            .with(MarkLogicSourceConfig.CONNECTION_PORT, testConfig.getRestPort())
+            .with(MarkLogicSourceConfig.CONNECTION_USERNAME, "admin")
```
rjrudin, line 63: Can get username/password from `testConfig` too

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConnector.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void stop() {
+        logger.info("Stopping MarkLogicSourceConnector");
+    }
+
+    @Override
+    public Class<? extends Task> taskClass() {
+        Class<? extends Task> clazz = RowBatcherSourceTask.class;
+        LoggerFactory.getLogger(getClass()).info("Task class: " + clazz);
```
rjrudin, line 43: This logging is only for showing whether DMSDK or Bulk DS is being used by the sink connector; I don't think it adds any value here since we'll only have one source task, so I don't think it's worth logging

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
BillFarber: I'll do this for the next PR.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
BillFarber: > We could reuse ml.dmsdk.batchSize, since RowBatcher is part of DMSDK
I don't particularly like the idea of prefixing all source-specific properties with ml.source.optic.

If you want them all (currently just 2) to start the same way, I'd go with ml.source.*

src/test/java/com/marklogic/kafka/connect/source/WriteToKafkaTest.java
BillFarber: I made it "ReadRowsViaOpticDslKafkaTest" based on the initial kafka-junit test PR where we agreed to name these tests "*KafkaTest".

PR #90, opened 2022-11-22
DEVEXP-108: Read rows via Optic DSL statement

BillFarber:
Part 2 - RowBatcher is pulling records from MarkLogic.


BillFarber:
I still need to abstract out a MarkLogicConfig class (from MarkLogicSinkConfig and MarkLogicSourceConfig) as well as a DefaultDatabaseClientConfigBuilder (from DefaultSinkDatabaseClientConfigBuilder and DefaultSourceDatabaseClientConfigBuilder). I attempted to do that for this PR, but I hit a wall. I will try again and either update this PR or create a new one.

BillFarber:
Something weird in the validator for ml.source.optic.dsl. If you run ReadRowsViaOpticDslKafkaTest, everything works fine.
If you change the ConfigDef define() for that property (line 40 in MarkLogicSourceConfig) , so that the default is null instead of "", then the connector does not load in that test.

src/test/java/com/marklogic/kafka/connect/source/RowBatcherSourceTaskTest.java
```diff
@@ -0,0 +1,52 @@
...
+    void setup() {
+        loadMarkLogicTestData();
+    }
+
+    @Test
+    @DisplayName("test the RowBatcher Source Task")
+    void testRowBatcherTask() {
+
+        RowBatcherSourceTask task = startSourceTask(
+            MarkLogicSourceConfig.CONNECTION_HOST, testConfig.getHost(),
```
rjrudin, line 27: The connection props aren't needed because `startSourceTask` is providing them.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
```diff
@@ -113,7 +113,7 @@ private SSLContext buildTwoWaySslContext(String certFile, String certPassword, M
 
         KeyManagerFactory keyManagerFactory;
         SSLContext sslContext;
-        try (InputStream keystoreInputStream = new FileInputStream(certFile)) {
+        try (InputStream keystoreInputStream = Files.newInputStream(Paths.get(certFile))) {
```
rjrudin, line 116: Do you know if this is fixing a bug, or just adopting a better practice for accessing the file?

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -33,20 +51,79 @@ public String version() {
...
+    protected RowBatcher<JsonNode> getNewRowBatcher() {
+        rowBatcher =  dataMovementManager.newRowBatcher(jsonHandle);
+        configureRowBatcher(parsedConfig, rowBatcher);
+        return rowBatcher;
+    }
+
+    protected void performPoll() {
+        dataMovementManager.startJob(rowBatcher);
+        rowBatcher.awaitCompletion();
+        rowBatcher = null;
```
rjrudin, line 81: This should call `stopJob` after `awaitCompletion` (we don't stop the job for writing because we want to keep the job going forever essentially, but a read job stops once all the batches have been processed). 

I say that only because the docs recommend doing that for QueryBatcher - https://docs.marklogic.com/guide/java/data-movement#id_77234 . I think DMSDK's intent is that you should still call `stopJob` so that the underlying implementation can do any cleanup that it would like to do. I checked the tests in the Java Client and we do have lots of RowBatcher tests that only call `awaitCompletion` and not `stopJob` afterwards, but I think it's better practice to call `stopJob` as well.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -15,10 +27,16 @@ public class RowBatcherSourceTask extends SourceTask {
 
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
+    private DatabaseClient databaseClient;
+    private DataMovementManager dataMovementManager;
+    private RowBatcher<JsonNode> rowBatcher = null;
+    private final ContentHandle<JsonNode> jsonHandle = new JacksonHandle().withFormat(Format.JSON).withMimetype("application/json");
```
rjrudin, line 33: Making this a final field makes me nervous - I don't think the Java Handle classes provide any assurance that they can be reused. It's common practice to create a new one each time you need one, and they're very lightweight, so I think we're better off just creating this each time we need a new RowBatcher.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -33,20 +51,79 @@ public String version() {
...
+        dataMovementManager.startJob(rowBatcher);
+        rowBatcher.awaitCompletion();
+        rowBatcher = null;
+    }
+
     @Override
     public void stop() {
-        logger.info("Stopping RowBatcherSourceTask");
+        if (rowBatcher != null) {
+            rowBatcher.awaitCompletion();
```
rjrudin, line 87: Per the javadocs for the `stop()` method in the superclass, I don't think we want to wait for completion - the user intent is to interrupt anything going on. So I think we just want `stopJob`, which will achieve the desired effect here.

rjrudin: I think you can test this as well. You can add a custom listener that sleeps for 500ms or so, and then in the test, immediately call "stop" after calling "poll".

rjrudin:
Found "NO_CONFIG_DEF", which solves the problem of a validator failing on a default value:

```
.define(DSL_PLAN, Type.STRING, ConfigDef.NO_DEFAULT_VALUE, new FreeFormStringValidator(), Importance.HIGH,
                "Required; a MarkLogic Optic DSL plan for querying the database.")
```

src/test/java/com/marklogic/kafka/connect/source/ReadRowsViaOpticDslKafkaTest.java
```diff
@@ -1,34 +1,35 @@
...
+    void setup() {
+        loadMarkLogicTestData();
+        setupKafka();
     }
 
     @Override
     // Until I start adding data, I don't want to delete any data
     protected String getJavascriptForDeletingDocumentsBeforeTestRuns() {
-        return "declareUpdate(); ";
+        return "declareUpdate(); cts.uris('', [], cts.trueQuery()).toArray().forEach(item => xdmp.documentDelete(item))";
```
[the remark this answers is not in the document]
BillFarber: Good point. Removed.

src/main/java/com/marklogic/kafka/connect/DefaultDatabaseClientConfigBuilder.java
BillFarber: I changed it because Jetbrains and Sonar recommended the change. Apparently FileInputStream implements a "finalize" method that can cause issues with garbage collection.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
BillFarber: I put a new test in. Take a look to see if it is what you meant.

rjrudin (changes requested):

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -1,77 +1,68 @@
...
+                "Sets the number of threads used for parallel reads from MarkLogic.");
+        MarkLogicConfig.addDefinitions(configDef);
+        return configDef;
+    }
+
+    private MarkLogicSourceConfig(final Map<?, ?> originals) {
         super(CONFIG_DEF, originals, false);
     }
 
+    static public class FreeFormStringValidator implements ConfigDef.Validator {
```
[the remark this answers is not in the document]
BillFarber: Fixed.
I realized the property name and value both get passed into the ensureValid method, so no new constructor is needed.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -1,77 +1,68 @@
...
-            "Sets the number of documents to be read in a batch from MarkLogic.")
-        .define(DMSDK_THREAD_COUNT, Type.INT, 8, Importance.MEDIUM,
-            "Sets the number of threads used for parallel reads from MarkLogic.");
+    public static final ConfigDef CONFIG_DEF = getConfigDef();
 
-    public MarkLogicSourceConfig(final Map<?, ?> originals) {
+    private static ConfigDef getConfigDef() {
+        ConfigDef configDef = new ConfigDef()
+            .define(DSL_PLAN, Type.STRING, ConfigDef.NO_DEFAULT_VALUE, new FreeFormStringValidator(), Importance.HIGH,
+                "Required; a MarkLogic Optic DSL plan for querying the database.")
```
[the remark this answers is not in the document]
BillFarber: Fixed

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -33,20 +51,79 @@ public String version() {
...
+        RowManager rowMgr = rowBatcher.getRowManager();
+        RawQueryDSLPlan plan = rowMgr.newRawQueryDSLPlan(new StringHandle(dslPlan));
+        rowBatcher.withBatchView(plan);
+
+        rowBatcher.onSuccess(event -> {
+            logger.debug("succeeded batch=" + event.getJobBatchNumber() +
+                " from " + event.getLowerBound() + " through " + event.getUpperBound() +
+                ((event.getJobBatchNumber() == 1) ? "\n" + event.getRowsDoc() : ""));
+            JsonNode rows = event.getRowsDoc().get("rows");
+            logger.info("JsonNode: \n" + rows.toPrettyString());
```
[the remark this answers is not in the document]
BillFarber: Correct.
However, I went ahead and removed the debug log since I'm not using it at all anymore.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -15,10 +27,16 @@ public class RowBatcherSourceTask extends SourceTask {
 
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
+    private DatabaseClient databaseClient;
+    private DataMovementManager dataMovementManager;
+    private RowBatcher<JsonNode> rowBatcher = null;
+    private ContentHandle<JsonNode> jsonHandle = new JacksonHandle().withFormat(Format.JSON).withMimetype("application/json");
```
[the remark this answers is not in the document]
BillFarber: I think I misunderstood your previous comment. I have moved it into the function that generates the RowBatcher instance.

src/test/java/com/marklogic/kafka/connect/sink/AbstractIntegrationSinkTest.java
```diff
@@ -25,26 +25,12 @@ public class AbstractIntegrationSinkTest extends AbstractSpringMarkLogicTest {
...
-
-    /**
-     * @param configParamNamesAndValues
+     * @param configParamNamesAndValues - Configuration values that need to be set for the test.
      * @return a MarkLogicSinkTask based on the default connection config and any optional config params provided by
      * the caller
      */
     protected AbstractSinkTask startSinkTask(String... configParamNamesAndValues) {
-        Map<String, String> config = newSinkConfig();
+        Map<String, String> config = newMarkLogicConfig(testConfig);
```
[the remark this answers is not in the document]
BillFarber: Fixed

src/test/java/com/marklogic/kafka/connect/source/RowBatcherSourceTaskTest.java
```diff
@@ -0,0 +1,78 @@
...
+import com.marklogic.client.document.XMLDocumentManager;
+import com.marklogic.client.io.FileHandle;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.util.concurrent.atomic.AtomicReference;
+
+class RowBatcherSourceTaskTest extends AbstractIntegrationSourceTest {
```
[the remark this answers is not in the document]
BillFarber: Changed.

PR #97, opened 2022-12-02
DEVEXP-106: Send failed records to dead letter queue when using DMSDK

BillFarber:


src/test/java/com/marklogic/kafka/connect/sink/SendWriteFailureRecordsToDlqKafkaTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @AfterEach
+    void tearDownKafka() {
+        kafka.stop();
+    }
+
+    @Test
+    void failedBatchesShouldGoToTheDlq() throws InterruptedException {
+        sendSomeJsonMessages(NUM_RECORDS);
```
rjrudin, line 55: I think it'd be good for this test to send a batch containing one valid document and one invalid document. And then we verify that neither was written to MarkLogic, and the one Throwable from the failed batch ended up in the DLQ... which makes me wonder, should the URIs of that failed batch going into the message that goes to the DLQ, like as part of the headers? That would provide very good visibility as to what failed.

src/main/java/com/marklogic/kafka/connect/sink/WriteBatcherSinkTask.java
```diff
@@ -58,15 +66,22 @@ public void put(Collection<SinkRecord> records) {
 
     @Override
     protected void writeSinkRecord(SinkRecord sinkRecord) {
-        this.writeBatcher.add(this.sinkRecordConverter.convert(sinkRecord));
+        try {
+            this.writeBatcher.add(this.sinkRecordConverter.convert(sinkRecord));
+        } catch (Exception e) {
+            sinkRecord.headers().addString("Failure type", "Record conversion");
```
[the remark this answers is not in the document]
BillFarber: "Failure type" is something I made up.
Part of the strategy of the DLQ is to easily be able to triage messages. If the record failed during the write process, then we can probably easily retry. If the record failed during the conversion, then we probably need to take a closer look.
So, my intention here is to give the hypothetical DLQ processor some information.

src/test/java/com/marklogic/kafka/connect/sink/SendConversionFailureRecordsToDlqKafkaTest.java
```diff
@@ -0,0 +1,134 @@
...
+    void tearDownKafka() {
+        kafka.stop();
+    }
+
+    @Test
+    void badMessagesShouldGoToTheDlq() throws InterruptedException {
+        Integer NUM_RECORDS = 2;
+        Integer NUM_BAD_RECORDS = 1;
+        sendSomeJsonMessages(NUM_RECORDS);
+        sendSomeBadJsonMessages(NUM_BAD_RECORDS);
```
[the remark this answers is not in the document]
BillFarber: 1) I changed the batchSize to 100 and the threadCount to 1, so that all the records would go at once.
2) However, the conversion error (at least the way I'm currently testing) is handled in Kafka code, not in my try-catch.

src/main/java/com/marklogic/kafka/connect/sink/WriteBatcherSinkTask.java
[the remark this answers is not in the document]
rjrudin: Sounds good - based on our experience with the NiFi connector, where the set of attributes is similar to the set of headers on a SourceRecord, let's give this a "marklogic-" prefix to make it clear that we own this - e.g. "marklogic-failure-type".

src/test/java/com/marklogic/kafka/connect/sink/SendWriteFailureRecordsToDlqKafkaTest.java
```diff
@@ -0,0 +1,111 @@
...
+        List<KeyValue<String, String>> records = new ArrayList<>();
+        for (int i = 0; i < numberOfRecords; i++) {
+            records.add(new KeyValue<>("aggregate", "{\"A\": \"" + i + "\"}"));
+        }
+        kafka.send(SendKeyValues.to(TOPIC, records));
+    }
+
+    private void assertCanConsumeRecordsFromTheTopic(String topic, Integer numRecords, String message) throws InterruptedException {
+        List<KeyValue<String, String>> consumedRecords = kafka.read(ReadKeyValues.from(topic).build());
+        assertEquals(numRecords, consumedRecords.size(), message);
```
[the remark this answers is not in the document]
BillFarber: I missed this.
Turns out adding headers was more complicated that I realized.
Fixed now, with assertions.

src/test/java/com/marklogic/kafka/connect/sink/SendWriteFailureRecordsToDlqKafkaTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private void sendSomeJsonMessages(Integer numberOfRecords) throws InterruptedException {
+        List<KeyValue<String, String>> records = new ArrayList<>();
+        for (int i = 0; i < numberOfRecords; i++) {
+            records.add(new KeyValue<>("aggregate", "{\"A\": \"" + i + "\"}"));
+        }
+        kafka.send(SendKeyValues.to(TOPIC, records));
+    }
+
+    private void assertMessageOnDlqAndHasExpectedHeaders(String topic, Integer numRecords, String message) throws InterruptedException {
```
rjrudin, line 96: Intellij shows numRecords and message as both being unused

src/test/java/com/marklogic/kafka/connect/sink/SendConversionFailureRecordsToDlqKafkaTest.java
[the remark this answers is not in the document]
rjrudin: Hmm - this test is seeming superfluous now, given that our connector only gets the 4 valid ones. I am thinking we should just drop this test - it would pass if we ran it before adding this story, so it's not verifying any new behavior. It's really only verifying that Kafka works. 

I think the `FailedConversionErrorReporterTest` is sufficient for us.

src/test/java/com/marklogic/kafka/connect/sink/HandleWriteFailureTest.java
```diff
@@ -4,16 +4,21 @@
...
 import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 
 public class HandleWriteFailureTest extends AbstractIntegrationSinkTest {
 
+    Throwable reportedException = null;
+    SinkRecord reportedSinkRecord = null;
```
rjrudin, line 20: We need one more SinkRecord now in this test - a valid one that has valid JSON in it. And we then want to make sure that 2 SinkRecords are sent to the ERR, as we don't know which one caused the failure.

src/main/java/com/marklogic/kafka/connect/sink/WriteBatcherSinkTask.java
```diff
@@ -215,6 +269,26 @@ public void processFailure(WriteBatch batch, Throwable throwable) {
                             values.get("kafka-offset"), values.get("kafka-timestamp"), values.get("kafka-topic"));
                     }
                 }
+                if (writeHandle instanceof SinkRecordMetadataHandle) {
+                    SinkRecord sinkRecord = ((SinkRecordMetadataHandle) writeHandle).getMySinkRecord();
+                    if (sinkRecord instanceof InternalSinkRecord) {
```
rjrudin, line 274: This looks identical to the block for conversion errors - should make this into a new private method to avoid the duplication.

src/test/java/com/marklogic/kafka/connect/sink/HandleWriteFailureTest.java
```diff
@@ -57,6 +63,19 @@ void handleWriteFailure() {
         assertEquals(key, values.get("kafka-key"));
         assertEquals(offset, Long.parseLong(values.get("kafka-offset")));
         assertEquals(timestamp, Long.parseLong(values.get("kafka-timestamp")));
+
+        Assertions.assertNotNull(reportedException,
+            "The mock error reporter should have been called with an exception");
+        Assertions.assertEquals(reportedException.getMessage(),
+            "Local message: failed to apply resource at documents: Bad Request. Server Message: XDMP-JSONDOC: xdmp:get-request-part-body(\"json\") -- Document is not JSON",
+            "The reported exception message does not match the expected message");
+        AtomicReference<Boolean> expectedFailureHeaderFound = new AtomicReference<>(false);
```
rjrudin, line 72: Same as for the FailedConversion test, should verify all 3 headers exist.

src/test/java/com/marklogic/kafka/connect/sink/AbstractIntegrationSinkTest.java
```diff
@@ -81,4 +100,8 @@ protected final void retryIfNotSuccessful(Runnable r, long sleepTime, int attemp
             }
         }
     }
+
+    public Map<String, Object> getTaskConfig() {
```
rjrudin, line 104: This is fine to have - I was thinking that since `taskConfig` is protected, a subclass could just directly access it. I do that a lot in tests, though much less often in application code. I think you can make `taskConfig` private as a result, and this can also be protected so that there's no implication that anything besides a subclass would use this. But it's fine the way it is too.

src/test/java/com/marklogic/kafka/connect/sink/FailedConversionErrorReporterTest.java
```diff
@@ -0,0 +1,62 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class FailedConversionErrorReporterTest extends AbstractIntegrationSinkTest {
+
+    Throwable reportedException = null;
+    SinkRecord reportedSinkRecord = null;
+    final private String someValidJsonToSignalAContrivedException = "{ \"A\": \"a\"}";
+
+    @Test
+    void sinkRecordWithNullValueCausesCallToErrorReporter() {
```
rjrudin, line 20: Just need one more assertion here - that the good record is written. To make that easy, can do the following:

1. Add e.g. `MarkLogicSinkConfig.DOCUMENT_COLLECTIONS, "good-data"` to the task config
2. Add `assertCollectionSize("good-data", 1)` to the end of the test, or rather after the flush occurs

rjrudin (changes requested):
LGTM, just one tiny assertion to add

src/test/java/com/marklogic/kafka/connect/sink/FailedConversionErrorReporterTest.java
BillFarber: drat, forgot this one.

PR #99, opened 2022-12-09
DEVEXP-110: Log errors when reading rows via Optic

BillFarber:


BillFarber:
I've been able to generate most of these messages on CP.

src/test/java/com/marklogic/kafka/connect/source/ReadRowsViaOpticDslTest.java
```diff
@@ -46,7 +46,8 @@ void noRowsReturned() throws InterruptedException {
             "expects no data to be returned.");
     }
 
-    @Test
+//    @Test
+// Not sure if we can test this now
     void testDslThatDoesNotStartWithFromView() {
```
rjrudin, line 51: I'm thinking that in the case of a bad config - i.e. a bad Optic plan - we should still fail fast on the `start()` method. So I'm thinking we should call `getNewRowBatcher` in `start()` to ensure that it can least be created, and then call `stopJob` on it right away. Though it looks like it should still check for the "known server bug" (when no rows exist - that shouldn't throw an error). 

Also, now that I say that above - for consistency, just do `newRowBatcher` for that method name, which is consistent at least with stuff I've been doing in the Java Client and some other projects. I like to reserve `get` for pure getters that just retrieve a value.

rjrudin (changes requested):
Looks good, just the one change to force `start()` to fail if the RowBatcher cannot be created due to a bad Optic plan. 

I verified too in CP that if poll() throws an exception, the connector stops immediately. So this seems like the right approach. 

Note that if the RowBatcher cannot be created in poll(), I could see an argument for throwing that exception in the future. But if we could create it in start(), that means the initial config was correct. So if it fails to be created in poll(), it seems that'd likely be due to an environment error. So I think logging that error and returning null makes good sense, but perhaps we'll do an option in the future to allow for the connector to fail in that scenario.

src/test/java/com/marklogic/kafka/connect/source/ReadRowsViaOpticDslTest.java
BillFarber: RowBatcher is not instantiated until poll() is called. So, we don't know if the Optic plan is bad until that point.
The call to start() doesn't even fail when I set the MarkLogic port incorrectly. Building the DatabaseClient must not actually talk to the database.

BillFarber: getNewRowBatcher changed to newRowBatcher

rjrudin: We agreed to omit this for now, as ML may simply not be available during start(), which could be acceptable

src/test/java/com/marklogic/kafka/connect/source/ReadRowsViaOpticDslTest.java
```diff
@@ -46,7 +46,8 @@ void noRowsReturned() throws InterruptedException {
             "expects no data to be returned.");
     }
 
-    @Test
+//    @Test
+// Not sure if we can test this now
```
rjrudin, line 50: As noted, can just change this to verify that null is returned by poll() instead of an exception

src/test/java/com/marklogic/kafka/connect/source/ReadRowsViaOpticDslTest.java
BillFarber: test changed to verify that when the DSL is invalid, poll() returns null

rjrudin (approved):

PR #100, opened 2022-12-12
DEVEXP-111: Read rows via Optic serialized plan

BillFarber:


src/test/java/com/marklogic/kafka/connect/source/ReadRowsViaOpticSerializedTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ReadRowsViaOpticSerializedTest extends AbstractIntegrationSourceTest {
+
+    @Test
+    void testRowBatcherTask() throws InterruptedException {
+        loadMarkLogicTestData();
```
rjrudin, line 21: Because this is almost identical to the test for a DSL query, I think the following can be used to avoid the duplication:

1. Make a protected method in the parent class of `verifyQueryReturnsFifteenAuthors` (as that captures explicitly what we expect to get back)
2. Move the helper methods into the parent class as well 

Arguably, you could make a separate class too - e.g. AuthorViewTester or something like that - so that the parent class  doesn't get stuffed with a bunch of helper methods only needed by a couple tests. That's a problem I've run into before with abstract parent test classes. But I think we can cross that bridge later. 

One other rename - `loadMarkLogicTestData` -> `loadFifteenAuthors`, which further makes it explicit what's going on in this scenario (or `loadFifteenAuthorsIntoMarkLogic` in case you think it's ambiguous whether "load" refers to ML or Kafka).

src/test/java/com/marklogic/kafka/connect/source/ReadRowsViaOpticSerializedTest.java
```diff
@@ -0,0 +1,71 @@
...
+        );
+
+        List<SourceRecord> newSourceRecords = task.poll();
+
+        Assertions.assertEquals(15, newSourceRecords.size());
+        assertTopicAndSingleValue(newSourceRecords, AUTHORS_TOPIC);
+    }
+
+    @Test
+    void noRowsReturned() throws InterruptedException {
```
rjrudin, line 36: I think this test is simple enough to keep it as-is, no need to bubble anything up to the parent class.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -145,10 +148,18 @@ private void configureRowBatcher(Map<String, Object> parsedConfig, RowBatcher<Js
         }
 
 
-        String dslPlan = (String) parsedConfig.get(MarkLogicSourceConfig.DSL_PLAN);
-        RowManager rowMgr = rowBatcher.getRowManager();
-        RawQueryDSLPlan plan = rowMgr.newRawQueryDSLPlan(new StringHandle(dslPlan));
-        rowBatcher.withBatchView(plan);
+        Boolean configuredForDsl = checkForNonEmptyString((String) parsedConfig.get(MarkLogicSourceConfig.DSL_QUERY));
```
rjrudin, line 151: To keep `configureRowBatcher` to a manageable size, move this to a `private void configureBatchView(rowBatcher, parsedConfig)` method - which could be static too, if you want to start trying out that pattern to see how it looks.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConnector.java
```diff
@@ -28,9 +31,24 @@ public ConfigDef config() {
...
+        if ((!(configuredForDsl || configuredForSerialized)) || (configuredForDsl && configuredForSerialized)) {
+            throw new ConfigException(
+                format("Either a DSL Optic query (%s) or a serialized Optic query (%s), but not not both, are required",
+                    MarkLogicSourceConfig.DSL_QUERY, MarkLogicSourceConfig.SERIALIZED_QUERY)
+            );
+        }
         this.config = config;
     }
 
+     static Boolean checkForNonEmptyString(String value) {
```
rjrudin, line 45: We have a StringUtils on the classpath already - it's the one from Spring via the ml-javaclient-util dependency - so can just use `StringUtils.hasText` instead of creating this function (it's fine if we depend on StringUtils.hasText in many places).

rjrudin (changes requested):
Looks good, just a few refactoring changes proposed.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
BillFarber: Is this an accidental repeat, or is there a difference between this one and the 2nd conversation from the top?

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConnector.java
BillFarber: darn it, I knew that existed, but I couldn't think of what library it was in and I couldn't find it. Thanks.

rjrudin (approved):
LGTM

PR #101, opened 2022-12-12
Prototyping functionality to save & retrieve connector state on a Kafka topic.

BillFarber:
* This is a not intended to be merged into mainline.

This is simply intended to prototype one possible method of saving and retrieving state information using a Kafka topic, from within a Kafka connector task.

The easiest way to see this in action, is to have CP running and then run the testRowBatcherTask test.
After performing the poll task logic, this will automatically create a topic in CP (stateTopic) and add a message with "state" data.
Additionally, on start, the task will get the more recent message from the topic (if one is available).


BillFarber:
Pushed in the feature/saveState branch for reference in the future.

PR #102, opened 2022-12-13
DEVEXP-112: Read rows as XML or CSV

BillFarber:
I'm considering creating an interface for the different types.
The implementations would replace some code in newRowBatcher() and also the onSuccess events.

### DIFF
--- a/src/main/java/com/marklogic/kafka/connect/source/CsvRowBatcherBuilder.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/CsvRowBatcherBuilder.java
@@ -1,0 +1,46 @@
+package com.marklogic.kafka.connect.source;
+
+import com.marklogic.client.datamovement.DataMovementManager;
+import com.marklogic.client.datamovement.RowBatchSuccessListener;
+import com.marklogic.client.datamovement.RowBatcher;
+import com.marklogic.client.io.Format;
+import com.marklogic.client.io.StringHandle;
+import com.marklogic.client.io.marker.ContentHandle;
+import org.apache.kafka.connect.source.SourceRecord;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
+import java.util.List;
+import java.util.Map;
+
+public class CsvRowBatcherBuilder extends AbstractRowBatchBuilder implements RowBatcherBuilder<String> {
+
+    CsvRowBatcherBuilder(DataMovementManager dataMovementManager, Map<String, Object> parsedConfig) {
+        super(dataMovementManager, parsedConfig);
+    }
+
+    public RowBatcher<String> newRowBatcher(List<SourceRecord> newSourceRecords) {
+        ContentHandle<String> contentHandle = new StringHandle().withFormat(Format.TEXT).withMimetype("text/csv");
+        RowBatcher<String> rowBatcher =  dataMovementManager.newRowBatcher(contentHandle);
+        configureRowBatcher(parsedConfig, rowBatcher);
+        rowBatcher.onSuccess(event -> onSuccessHandler(event, newSourceRecords));
+        return rowBatcher;
+    }
+
+    protected void onSuccessHandler(RowBatchSuccessListener.RowBatchResponseEvent<String> event, List<SourceRecord> newSourceRecords) {
+        String document = event.getRowsDoc();
+        logger.debug("CSV document: \n" + document);
+        BufferedReader reader = new BufferedReader(new StringReader(document));
+        try {
+            String headers = reader.readLine();
+            reader.lines().forEach(line -> {
+                String newDocument = headers + "\n" + line;
+                SourceRecord newRecord = new SourceRecord(null, null, topic, null, newDocument);
+                newSourceRecords.add(newRecord);
+            });
+        } catch (IOException ex) {
+            logBatchError(ex);
+        }
+    }
+}
--- a/src/main/java/com/marklogic/kafka/connect/source/JsonRowBatcherBuilder.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/JsonRowBatcherBuilder.java
@@ -1,0 +1,48 @@
+package com.marklogic.kafka.connect.source;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.client.datamovement.DataMovementManager;
+import com.marklogic.client.datamovement.RowBatchSuccessListener;
+import com.marklogic.client.datamovement.RowBatcher;
+import com.marklogic.client.io.Format;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.io.marker.ContentHandle;
+import org.apache.kafka.connect.source.SourceRecord;
+
+import java.util.List;
+import java.util.Map;
+
+public class JsonRowBatcherBuilder extends AbstractRowBatchBuilder implements RowBatcherBuilder<JsonNode> {
+
+    JsonRowBatcherBuilder(DataMovementManager dataMovementManager, Map<String, Object> parsedConfig) {
+        super(dataMovementManager, parsedConfig);
+    }
+
+    public RowBatcher<JsonNode> newRowBatcher(List<SourceRecord> newSourceRecords) {
+        ContentHandle<JsonNode> contentHandle = new JacksonHandle().withFormat(Format.JSON).withMimetype("application/json");
+        RowBatcher<JsonNode> rowBatcher =  dataMovementManager.newRowBatcher(contentHandle);
+        configureRowBatcher(parsedConfig, rowBatcher);
+        rowBatcher.onSuccess(event -> onSuccessHandler(event, newSourceRecords));
+        return rowBatcher;
+    }
+
+    protected void onSuccessHandler(RowBatchSuccessListener.RowBatchResponseEvent<JsonNode> event, List<SourceRecord> newSourceRecords) {
+        JsonNode rows = event.getRowsDoc().get("rows");
+        logger.debug("JsonNode: \n" + rows.toPrettyString());
+
+        for (JsonNode row : rows) {
+// We may need to add a switch to include a key in the record depending on how the target topic is configured.
+// If the topic's cleanup policy is set to "compact", then a key is required to be included in the SourceRecord.
+//            String key = event.getJobBatchNumber() + "-" + rowNumber;
+
+            // Calling toString on the JsonNode, as when this is run in Confluent Platform, we get the following
+            // error: org.apache.kafka.connect.errors.DataException: Java class com.fasterxml.jackson.databind.node.ObjectNode does not have corresponding schema type.
+            try {
+                SourceRecord newRecord = new SourceRecord(null, null, topic, null, row.toString());
+                newSourceRecords.add(newRecord);
+            } catch (Exception ex) {
+                logBatchError(ex, row.toString());
+            }
+        }
+    }
+}
--- a/src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
@@ -1,24 +1,15 @@
 package com.marklogic.kafka.connect.source;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.FailedRequestException;
 import com.marklogic.client.datamovement.*;
 import com.marklogic.client.ext.DatabaseClientConfig;
 import com.marklogic.client.ext.DefaultConfiguredDatabaseClientFactory;
-import com.marklogic.client.io.Format;
-import com.marklogic.client.io.JacksonHandle;
-import com.marklogic.client.io.StringHandle;
-import com.marklogic.client.io.marker.ContentHandle;
-import com.marklogic.client.row.RawPlanDefinition;
-import com.marklogic.client.row.RawQueryDSLPlan;
-import com.marklogic.client.row.RowManager;
 import com.marklogic.kafka.connect.DefaultDatabaseClientConfigBuilder;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.apache.kafka.connect.source.SourceTask;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.util.StringUtils;
 
 import java.util.*;
 
@@ -31,9 +22,9 @@ public class RowBatcherSourceTask extends SourceTask {
 
     private DatabaseClient databaseClient;
     private DataMovementManager dataMovementManager;
-    private RowBatcher<JsonNode> rowBatcher = null;
-    private Map<String, Object> parsedConfig;
+    private RowBatcher<?> rowBatcher = null;
     private long pollDelayMs = 1000L;
+    private RowBatcherBuilder<?> rowBatcherBuilder;
 
     /**
      * Required for a Kafka task.
@@ -53,11 +44,13 @@ public class RowBatcherSourceTask extends SourceTask {
     @Override
     public final void start(Map<String, String> config) {
         logger.debug("Starting RowBatcherSourceTask");
-        this.parsedConfig = MarkLogicSourceConfig.CONFIG_DEF.parse(config);
+        Map<String, Object> parsedConfig = MarkLogicSourceConfig.CONFIG_DEF.parse(config);
         DatabaseClientConfig databaseClientConfig = new DefaultDatabaseClientConfigBuilder().buildDatabaseClientConfig(parsedConfig);
         this.databaseClient = new DefaultConfiguredDatabaseClientFactory().newDatabaseClient(databaseClientConfig);
         dataMovementManager = databaseClient.newDataMovementManager();
         pollDelayMs = (Long) parsedConfig.get(MarkLogicSourceConfig.WAIT_TIME);
+
+        rowBatcherBuilder = newRowBatcherBuilder(dataMovementManager, parsedConfig);
     }
 
     @Override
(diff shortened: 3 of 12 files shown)


src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
```diff
@@ -31,6 +31,21 @@ public class AbstractIntegrationSourceTest extends AbstractIntegrationTest {
     protected final String AUTHORS_OPTIC_DSL = "op.fromView(\"Medical\", \"Authors\")";
     protected final String AUTHORS_OPTIC_SERIALIZED = "{\"$optic\":{\"ns\":\"op\", \"fn\":\"operators\", \"args\":[{\"ns\":\"op\", \"fn\":\"from-view\", \"args\":[\"Medical\", \"Authors\"]}]}}";
     protected final String AUTHORS_TOPIC = "Authors";
+    protected final String JSON_RESULT = "{\"Medical.Authors.ID\":{\"type\":\"xs:integer\",\"value\":2}," +
+        "\"Medical.Authors.LastName\":{\"type\":\"xs:string\",\"value\":\"Pulhoster\"}," +
+        "\"Medical.Authors.ForeName\":{\"type\":\"xs:string\",\"value\":\"Misty\"}}";
+    protected final String XML_RESULT = "<Record><t:columns xmlns:t=\"http://marklogic.com/table\">\n" +
```
rjrudin, line 37: Just define the XML and CSV results in the classes that need them, which likely will only be those two classes. JSON result is fine to have here, although I'd probably stuff it in ReadRowsViaOpticDslTest as a public static and have the serialized test reference it there. That helps keep this parent class a little more lean.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSupplier.java
```diff
@@ -0,0 +1,73 @@
...
+import com.marklogic.client.row.RawQueryDSLPlan;
+import com.marklogic.client.row.RowManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.util.StringUtils;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+public abstract class RowBatcherSupplier implements Supplier<RowBatcher<?>> {
```
rjrudin, line 16: Fun convenience class - extending LoggingObject (from ml-javaclient-util), which declares a logger for you and also a `format` method so you don't have to type `String.format`. Then you don't have to declare logger here or in its subclasses.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -82,11 +74,23 @@ public List<SourceRecord> poll() throws InterruptedException {
...
+                rowBatcherSupplier = new JsonRowBatcherSupplier(dataMovementManager, newSourceRecords, parsedConfig);
+                break;
+            case XML:
+                rowBatcherSupplier = new XmlRowBatcherSupplier(dataMovementManager, newSourceRecords, parsedConfig);
+                break;
+            case CSV:
+                rowBatcherSupplier = new CsvRowBatcherSupplier(dataMovementManager, newSourceRecords, parsedConfig);
+                break;
+            default:
+                throw new Exception("");
```
rjrudin, line 91: We don't expect this, but in the exceedingly rare event that it does happen, a nice message is still helpful - I would do `throw new IllegalArgumentException("Unexpected output type: " + outputType);` . That also extend RuntimeException so you don't have to declare `throws Exception` on this method.

src/main/java/com/marklogic/kafka/connect/source/JsonRowBatcherSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+    private final List<SourceRecord> sourceRecords;
+
+    JsonRowBatcherSupplier(DataMovementManager dataMovementManager, List<SourceRecord> sourceRecords, Map<String, Object> parsedConfig) {
+        this.dataMovementManager = dataMovementManager;
+        this.sourceRecords = sourceRecords;
+        this.parsedConfig = parsedConfig;
+        this.topic = (String) parsedConfig.get(MarkLogicSourceConfig.TOPIC);
+    }
+
+    public RowBatcher<JsonNode> get() {
```
rjrudin, line 31: Seeing this, I think I gave bad advice with recommending the use of Supplier. This of course works fine, but I don't like how `get()` is ambiguous about whether it returns a new instance or the same instance - the javadocs for Supplier even state that it's intended to be ambiguous. And because `get()` is no-arg, it means that the supplier has to hold onto the List object as state. 

I think a `Function<List<SourceRecord>, RowBatcher>` is the better way to go - but that's a bit ugly, and the Function method of `apply` is not intuitive either. So I think a custom interface is better, like you had in mind originally:

```
interface RowBatcherBuilder 
  RowBatcher newRowBatcher(List<SourceRecord> list)
```

That has another benefit too - you can now look at output type in the `start()` method and instantiate a `RowBatcherBuilder` and reuse it on every call to `poll()`. The RBB will need the DataMovementManager and config Map passed to its constructor, and it can hold onto both of those as class fields because both are not intended to be modified. 

Then, `poll()` just does the following:

```
    List<SourceRecord> records = new ...
    RowBatcher rowBatcher = this.rowBatcherBuilder.newRowBatcher(records);
```

That makes it clear that a new list is created every time and it's used to create the RowBatcher, and thus there's no reason for the RBB to hold onto it as state. 

Sorry for the bad advice with Supplier, but I think this is a worthwhile refactor.

src/main/java/com/marklogic/kafka/connect/source/JsonRowBatcherSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+        ContentHandle<JsonNode> contentHandle = new JacksonHandle().withFormat(Format.JSON).withMimetype("application/json");
+        RowBatcher<JsonNode> rowBatcher =  dataMovementManager.newRowBatcher(contentHandle);
+        configureRowBatcher(parsedConfig, rowBatcher);
+        rowBatcher.onSuccess(event -> onSuccessHandler(event, sourceRecords));
+        return rowBatcher;
+    }
+
+    protected void onSuccessHandler(RowBatchSuccessListener.RowBatchResponseEvent<?> event, List<SourceRecord> newSourceRecords) {
+        @SuppressWarnings("unchecked") com.fasterxml.jackson.databind.JsonNode rows = ((RowBatchSuccessListener.RowBatchResponseEvent<com.fasterxml.jackson.databind.JsonNode>) event).getRowsDoc().get("rows");
```
rjrudin, line 40: You can avoid this casting by explicitly declaring `RowBatcherResponseEvent<JsonNode>` in the method signature.

src/main/java/com/marklogic/kafka/connect/source/XmlRowBatcherSupplier.java
```diff
@@ -0,0 +1,101 @@
...
+        Node columnsNode = docElement.getElementsByTagNameNS(TABLE_NS_URI, "columns").item(0);
+        NodeList rows = docElement.getElementsByTagNameNS(TABLE_NS_URI, "row");
+
+        for (int i=0; i < rows.getLength(); i++) {
+            try {
+                Document newDocument = convertRowNodeToNewDocument(columnsNode, rows.item(i));
+                SourceRecord newRecord = new SourceRecord(null, null, topic, null, documentToString(newDocument));
+                newSourceRecords.add(newRecord);
+            } catch (Exception ex) {
+                logger.error("Failed to create or add SourceRecord from result row; cause: " + ex.getMessage());
```
rjrudin, line 67: Just do one log message here - we want to treat this as one "log event"; you can just append the "failed result row" after the exception message.

src/main/java/com/marklogic/kafka/connect/source/XmlRowBatcherSupplier.java
```diff
@@ -0,0 +1,101 @@
...
+
+        newRoot.appendChild(newColumns);
+        newRoot.appendChild(newRow);
+        logger.debug("XML root: \n" + documentToString(newDocument));
+        return newDocument;
+    }
+
+    private static String documentToString(Document newDoc) throws TransformerException {
+        DOMSource domSource = new DOMSource(newDoc);
+        Transformer transformer = TransformerFactory.newInstance().newTransformer();
```
rjrudin, line 94: Should stash this Transformer as a class attribute during the constructor, there's a cost to creating both it and the TransformerFactory. Though - check the javadocs on whether a Transformer is threadsafe - it may be that we need to stash the TransformerFactory (which I'm almost certain is threadsafe) and create a new Transformer every time.

src/main/java/com/marklogic/kafka/connect/source/CsvRowBatcherSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+    public RowBatcher<String> get() {
+        ContentHandle<String> contentHandle = new StringHandle().withFormat(Format.TEXT).withMimetype("text/csv");
+        RowBatcher<String> rowBatcher =  dataMovementManager.newRowBatcher(contentHandle);
+        configureRowBatcher(parsedConfig, rowBatcher);
+        rowBatcher.onSuccess(event -> onSuccessHandler(event, sourceRecords));
+        return rowBatcher;
+    }
+
+    protected void onSuccessHandler(RowBatchSuccessListener.RowBatchResponseEvent<?> event, List<SourceRecord> newSourceRecords) {
+        @SuppressWarnings("unchecked") String document = ((RowBatchSuccessListener.RowBatchResponseEvent<String>) event).getRowsDoc();
```
rjrudin, line 42: Can declare "String" in the method signature to avoid this casting.

rjrudin (commented):
Looks good, except for my lousy advice about using a Supplier.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
BillFarber: Thanks, meant to come back to this, but forgot.

src/main/java/com/marklogic/kafka/connect/source/XmlRowBatcherSupplier.java
BillFarber: From the javadocs:
An object of this class may not be used in multiple threads running concurrently. Different Transformers may be used concurrently by different threads.

BillFarber: Since neither of those classes are thread safe, I moved the instantiations to the beginning of the onSuccess handler.

src/main/java/com/marklogic/kafka/connect/source/CsvRowBatcherBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+        BufferedReader reader = new BufferedReader(new StringReader(document));
+        try {
+            String headers = reader.readLine();
+            reader.lines().forEach(line -> {
+                String newDocument = headers + "\n" + line;
+                SourceRecord newRecord = new SourceRecord(null, null, topic, null, newDocument);
+                newSourceRecords.add(newRecord);
+            });
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
rjrudin, line 46: Java IO is annoying with its checked exceptions - we certainly don't expect this IOException here, but we need to catch it and log it just like in the JSON builder. I think a convenience method would be good here - I would add AbstractRowBatchBuilder as a parent class for each RBB impl and have a "logBatchError(Exception ex)" method that logs the message that the JSON builder is logging.

src/main/java/com/marklogic/kafka/connect/source/JsonRowBatcherBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+// If the topic's cleanup policy is set to "compact", then a key is required to be included in the SourceRecord.
+//            String key = event.getJobBatchNumber() + "-" + rowNumber;
+
+            // Calling toString on the JsonNode, as when this is run in Confluent Platform, we get the following
+            // error: org.apache.kafka.connect.errors.DataException: Java class com.fasterxml.jackson.databind.node.ObjectNode does not have corresponding schema type.
+            try {
+                SourceRecord newRecord = new SourceRecord(null, null, topic, null, row.toString());
+                newSourceRecords.add(newRecord);
+            } catch (Exception ex) {
+                logger.error("Failed to create or add SourceRecord from result row; cause: " + ex.getMessage());
```
rjrudin, line 47: Just do one logger.error call, can stuff everything together - having two log "events" for one error can make life difficult for users that e.g. send log messages to a tool like Grafana, where ideally an error is represented by a single log event (so the error doesn't get counted twice).

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -82,13 +81,31 @@ public List<SourceRecord> poll() throws InterruptedException {
...
-        ContentHandle<JsonNode> jsonHandle = new JacksonHandle().withFormat(Format.JSON).withMimetype("application/json");
-        rowBatcher =  dataMovementManager.newRowBatcher(jsonHandle);
-        configureRowBatcher(parsedConfig, rowBatcher, newSourceRecords);
+    protected RowBatcher<?> newRowBatcher(List<SourceRecord> newSourceRecords) {
+        rowBatcher = rowBatcherBuilder.newRowBatcher(newSourceRecords);
+        configureRowBatcher(parsedConfig, rowBatcher);
         return rowBatcher;
     }
 
+    protected RowBatcherBuilder<?> newRowBatcherBuilder(DataMovementManager dataMovementManager, Map<String, Object> parsedConfig) {
```
rjrudin, line 90: Make this and configureRowBatcher private - the benefit being that it makes it more obvious which methods are exposed for testing (the protected ones).

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -120,7 +137,7 @@ public synchronized void stop() {
      * @param parsedConfig - The complete configuration object including any transform parameters.
      * @param rowBatcher - The RowBatcher object to be configured.
      */
-    private void configureRowBatcher(Map<String, Object> parsedConfig, RowBatcher<JsonNode> rowBatcher, List<SourceRecord> newSourceRecords) {
+    void configureRowBatcher(Map<String, Object> parsedConfig, RowBatcher<?> rowBatcher) {
```
rjrudin, line 140: This is fine to keep here. I do think there's some benefit in having RowBatcherBuilder encapsulate all knowledge of how to construct and configure a RowBatcher so that the Task class has no understanding of how that works (I would do that by having a parent AbstractRBB class and tossing this method in there for each subclass to use). For example, we could write small unit tests that invoke an RBB and then inspect the RB that comes back to make sure it's correct. But we can cross that bridge later (if ever), this is fine here.

src/main/java/com/marklogic/kafka/connect/source/XmlRowBatcherBuilder.java
```diff
@@ -0,0 +1,97 @@
...
+        return rowBatcher;
+    }
+
+    protected void onSuccessHandler(RowBatchSuccessListener.RowBatchResponseEvent<Document> event, List<SourceRecord> newSourceRecords) {
+        TransformerFactory transformerFactory = TransformerFactory.newInstance();
+        Transformer transformer;
+        try {
+            transformer = transformerFactory.newTransformer();
+        } catch (TransformerConfigurationException e) {
+            throw new RuntimeException("Unable to create XML Transformer: " + e.getMessage());
```
rjrudin, line 50: Gotta catch and log this too. I think it's perfectly fine to just throw everything in here into one try/catch block, logging the generic error message I mentioned above. We really don't expect any errors here, so if there is an error, I think it's fine to just log it and keep going.

rjrudin (changes requested):
Looks good, just some changes requested for error handling and logging.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
BillFarber: Since I created the AbstractRBB above, I moved this (and others).

src/main/java/com/marklogic/kafka/connect/source/XmlRowBatcherBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            transformer = transformerFactory.newTransformer();
+            Document rowsDoc = event.getRowsDoc();
+
+            Element docElement = rowsDoc.getDocumentElement();
+            Node columnsNode = docElement.getElementsByTagNameNS(TABLE_NS_URI, "columns").item(0);
+            NodeList rows = docElement.getElementsByTagNameNS(TABLE_NS_URI, "row");
+
+            for (int i=0; i < rows.getLength(); i++) {
+                Document newDocument = convertRowNodeToNewDocument(columnsNode, rows.item(i), transformer);
```
rjrudin, line 53: Sorry, but I now realize we need two try/catch's for XML handling - 1) the overall one, to catch any JAXP issues, or possibly issues with obtaining the "row" and "columnsNode" nodes; and 2) a try/catch for every row in the `for` loop, similar to what's done for CSV/JSON. CSV/JSON don't need that first try/catch because we don't have any meaningful initialization to do first. 

Alternatively, you could move the code for doing the JAXP and XML parsing stuff into a private method that returns a Tuple<Node, Node> - the first Node being the columns, the second Node being the rows. And you could wrap that in a try/catch so that if we failed to get those two nodes, we log the error and don't do anything further. If we do get those nodes, then we hop into the for loop and do a try/catch on each row.

rjrudin (commented):
Sorry, one more error thing I realized we need for XML.

rjrudin (approved):

PR #105, opened 2022-12-19
DEVEXP-113: Constrain reading rows by user-defined column

BillFarber:
Not ready yet, but probably worth taking a look at to make sure we're going in the right direction.


src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
```diff
@@ -62,16 +65,20 @@ protected void configureRowBatcher(Map<String, Object> parsedConfig, RowBatcher<
...
     }
 
-    private void configureBatchView(Map<String, Object> parsedConfig, RowBatcher<?> rowBatcher) {
+    private void configureBatchView(Map<String, Object> parsedConfig, RowBatcher<?> rowBatcher, String previousBatchMaxValue) {
         boolean configuredForDsl = StringUtils.hasText((String) parsedConfig.get(MarkLogicSourceConfig.DSL_QUERY));
         if (configuredForDsl) {
-            String dslQuery = (String) parsedConfig.get(MarkLogicSourceConfig.DSL_QUERY);
+            currentQuery = (String) parsedConfig.get(MarkLogicSourceConfig.DSL_QUERY);
+            if (previousBatchMaxValue != null) {
+                currentQuery = injectConstraintIntoDslQuery(currentQuery, constrainingColumn, previousBatchMaxValue);
```
rjrudin, line 77: Ooh, I like "inject"

src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
```diff
@@ -84,4 +91,23 @@ private void configureBatchView(Map<String, Object> parsedConfig, RowBatcher<?>
     private void onFailureHandler(RowBatchFailureListener.RowBatchFailureEvent batch, Throwable throwable) {
         logger.error("batch "+batch.getJobBatchNumber()+" failed with error: "+throwable.getMessage());
     }
+
+    protected static String injectConstraintIntoDslQuery(String originalDsl, String constrainingColumn, String constrainingValue) {
+        String constrainingPhrase = ".where(op.gt(op.col(\"" + constrainingColumn + "\"), " + constrainingValue + "))";
```
rjrudin, line 96: I think we need a test for a date value, which I'm pretty sure will need to be quoted (and single quotes should work fine). And while I don't know the risks of "Optic injection attack", I think we'd want to err on the side of safety and remove any single/double quotes in both the column and the value - not that we expect them, but for safety reasons. 

A refactor to consider too for the serialized plan story (out of scope for now) - because you're unit testing this (good), and because the logic isn't trivial, I think making a "ConstraintInjector" interface with  "DslQueryConstraintInjector" and "SerializedQueryConstraintInjector" implementations would be good. That would increase readability and make unit testing easier, as you could test the logic in isolation without having to construct an RBB.

src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
```diff
@@ -16,11 +16,14 @@ public class AbstractRowBatchBuilder extends LoggingObject {
     protected final DataMovementManager dataMovementManager;
     protected final Map<String, Object> parsedConfig;
     protected final String topic;
+    protected final String constrainingColumn;
```
rjrudin, line 19: These two should be private since no one needs them (yet) except for this class.

src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
```diff
@@ -84,4 +91,23 @@ private void configureBatchView(Map<String, Object> parsedConfig, RowBatcher<?>
...
+            int firstClosingParen = originalDslNoWhitespace.indexOf(").");
+            constrainedDsl = originalDslNoWhitespace.substring(0, firstClosingParen+1)
+                + constrainingPhrase + originalDslNoWhitespace.substring(firstClosingParen+1);
+        } else {
+            constrainedDsl = originalDsl + constrainingPhrase;
+        }
+        return constrainedDsl;
+    }
+
+    public String getCurrentQuery() {
```
rjrudin, line 110: Instead of modifying the interface, what if `newRowBatcher` returned both the RowBatcher and the query? I've used `Pair` in the past for when exactly two things need to be returned, but a custom class - e.g. `RowBatcherContext` - would be useful, particularly when we need to return a third thing. 

I generally prefer that because it a) avoids additional state in the class, b) avoids another method in the interface, and c) most importantly, makes it clear to the reader that the RowBatcher and the query are constructed at the exact same time and are not mutated later on.

src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
```diff
@@ -37,7 +40,7 @@ void logBatchError(Exception ex) {
      * @param parsedConfig - The complete configuration object including any transform parameters.
      * @param rowBatcher - The RowBatcher object to be configured.
      */
-    protected void configureRowBatcher(Map<String, Object> parsedConfig, RowBatcher<?> rowBatcher) {
+    protected void configureRowBatcher(Map<String, Object> parsedConfig, RowBatcher<?> rowBatcher, String previousBatchMaxValue) {
```
rjrudin, line 43: I know this is verbose, but I think it's important that this param be clearly associated with the constraining column, so I'm thinking `previousMaxConstraintColumnValue`. Again, that's wordy, but I'm not sure how to shorten it (it's a simpler name if we assume like Kafka JDBC does that it's always a timestamp, but we know that won't necessarily be the case).

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -18,6 +18,7 @@ public class MarkLogicSourceConfig extends MarkLogicConfig {
 
     public static final String DSL_QUERY = "ml.source.optic.dsl";
     public static final String SERIALIZED_QUERY = "ml.source.optic.serialized";
+    public static final String CONSTRAINT_COLUMN_NAME = "ml.source.optic.constraintColumn.name";
```
rjrudin, line 21: Just realized there's both "constraintColumn" and "constrainingColumn" - I think the former is a little better, so use that throughout the PR.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -65,7 +77,7 @@ public List<SourceRecord> poll() throws InterruptedException {
 
         long start = System.currentTimeMillis();
         try {
-            rowBatcher = newRowBatcher(newSourceRecords);
+            rowBatcher = newRowBatcher(previousBatchMaxValue, newSourceRecords);
```
rjrudin, line 80: I think newSourceRecords should be first since it will never be null, while the previous value may be null. I generally prefer ordering params based on required stuff first, optional stuff second.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -140,4 +157,23 @@ private boolean rowBatcherErrorIsKnownServerBug(Exception ex) {
         }
         return false;
     }
+
+    private String buildMaxValueDslQuery(String currentQuery, String constrainingColumn) {
+        return String.format("%s.orderBy(op.desc(\"%s\")).limit(1).select([op.as(\"constraint\", op.col(\"%s\"))])", currentQuery, constrainingColumn, constrainingColumn);
```
rjrudin, line 162: Nice idea with the final `select`! I like how that simplifies it.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -140,4 +157,23 @@ private boolean rowBatcherErrorIsKnownServerBug(Exception ex) {
         }
         return false;
     }
+
+    private String buildMaxValueDslQuery(String currentQuery, String constrainingColumn) {
```
rjrudin, line 161: The complexity of this logic - and my experience with QueryMarkLogic in NiFi, where all of this complexity was stuffed into the QueryMarkLogic class (I've moved some of it out) and thus was fairly difficult to understand - makes me think the refactor I mentioned above will definitely be worth it. And instead of "ConstraintInjector", I think the interface would be for "I can both inject a constraint and also get you the max value for a constraint column", so.... perhaps `ConstraintColumnHandler` for lack of a better verb now (handle = process = "I'm not sure what else to call this"). 

And then I think two methods:

- `injectConstraintIntoQuery`
- `getMaxConstraintValue`

And then we'll have DslQuery and SerializedQuery implementations. 

Again, no need to do this now, as it'll be better to wait to do the serialized query implementation to make sure that the refactor makes sense.

src/test/java/com/marklogic/kafka/connect/source/ConstraintInjectionTest.java
```diff
@@ -0,0 +1,54 @@
+package com.marklogic.kafka.connect.source;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class ConstraintInjectionTest extends AbstractIntegrationSourceTest {
+
+    @Test
+    void testConstraintInjection() {
+        String originalDsl = "op.fromView(\"Medical\", \"Authors\")";
```
rjrudin, line 10: I think these should all be separate unit tests, each with a good name that identifies the scenario they're testing. That also allows each one to succeed/fail in isolation and makes each one easier to understand. 

Also, I like how the "max value" test is also exercising the query - I think that's necessary here too. It's good to verify that the query is exactly as we expect it, but the test should also execute the query to verify that we get back the expected number of rows.

src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
BillFarber: Deferred until the serialized constraint story.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
BillFarber: Deferred until the serialized constraint story.

src/test/java/com/marklogic/kafka/connect/source/ConstraintInjectionTest.java
BillFarber: Well, this sucks.
Limit doesn't work here. I'm assuming the problem is with RowBatcher.

src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
BillFarber: Tests in place for the first half.

src/main/java/com/marklogic/kafka/connect/source/AbstractRowBatchBuilder.java
```diff
@@ -82,6 +89,50 @@ private void configureBatchView(Map<String, Object> parsedConfig, RowBatcher<?>
...
+            int locationOfNextSpace = message.indexOf(" ", locationOfErrorString+lengthOfErrorString);
+            String columnName = message.substring(locationOfErrorString+lengthOfErrorString, locationOfNextSpace);
+            logger.error("batch "+batch.getJobBatchNumber()+" failed due to missing column: " + columnName);
+        } else {
+            logger.error("batch " + batch.getJobBatchNumber() + " failed with error: " + throwable.getMessage());
+        }
+    }
+
+    protected static String injectConstraintIntoDslQuery(String originalDsl, String constraintColumn, String constraintValue) {
+        String constraintPhrase = ".where(op.gt(op.col(\"" + constraintColumn + "\"), " + constraintValue + "))";
```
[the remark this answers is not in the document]
BillFarber: Added to my list for the 2nd PR for this story (primarily the serialized query part).

src/main/java/com/marklogic/kafka/connect/source/QueryContext.java
```diff
@@ -0,0 +1,34 @@
...
+        this.rowBatcher = rowBatcher;
+        this.currentQuery = currentQuery;
+        this.sourceRecords = sourceRecords;
+    }
+
+    public RowBatcher<?> getRowBatcher() {
+        return rowBatcher;
+    }
+
+    public String getCurrentQuery() {
```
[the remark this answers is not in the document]
BillFarber: The query here is either the original query, or an updated query with the constraint phrase injected.
I couldn't come up with a good name for that variable.

src/test/java/com/marklogic/kafka/connect/source/ConstraintInjectionTest.java
BillFarber: Resolved with: MarkLogicSourceConfig.DMSDK_BATCH_SIZE = String.valueOf(Integer.MAX_VALUE)

PR #106, opened 2022-12-19
DEVEXP-111: Read rows via Optic serialized plan

BillFarber:
This fixes a bug where the Validator was forcing a value for the DSL and Serialized plans.

### DIFF
--- a/src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
@@ -36,9 +36,9 @@ public class MarkLogicSourceConfig extends MarkLogicConfig {
         ConfigDef configDef = new ConfigDef();
         MarkLogicConfig.addDefinitions(configDef);
         return configDef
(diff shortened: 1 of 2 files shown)


rjrudin (approved):
Just to confirm - we can't do any validation here because a Kafka validator is limited to only seeing one property and its value? I'm wondering if there's a place for a check later on so that things blow up if neither is specified?

BillFarber:
> Just to confirm - we can't do any validation here because a Kafka validator is limited to only seeing one property and its value? I'm wondering if there's a place for a check later on so that things blow up if neither is specified?

That is correct.
So, now when you start the connector, we detect if neither or both are filled in. In that case, the connector throws an exception with a message: 

[2022-12-19 15:56:39,476] ERROR WorkerConnector{id=MarkLogicSourceConnectorConnector_0} Error while starting connector (org.apache.kafka.connect.runtime.WorkerConnector)
org.apache.kafka.common.config.ConfigException: Either a DSL Optic query (ml.source.optic.dsl) or a serialized Optic query (ml.source.optic.serialized), but not both, are required
    at com.marklogic.kafka.connect.source.MarkLogicSourceConnector.start(MarkLogicSourceConnector.java:39)

PR #107, opened 2022-12-22
DEVEXP-113: Constrain reading rows by user-defined column, serialized query

BillFarber:
DEVEXP-113: Constrain reading rows by user-defined column, serialized query


src/main/java/com/marklogic/kafka/connect/source/ConstraintColumnHandler.java
```diff
@@ -0,0 +1,11 @@
+package com.marklogic.kafka.connect.source;
+
+import com.marklogic.client.datamovement.RowBatcher;
+
+import java.util.Map;
+
+public interface ConstraintColumnHandler {
+    String injectConstraintIntoQuery(Map<String, Object> parsedConfig, String constraintColumn, String constraintValue);
+    String getMaxConstraintValue(String currentQuery, long queryStartTimeInMillis, String constraintColumn);
+    void addQueryToRowBatcher(String currentQuery, RowBatcher<?> rowBatcher);
```
rjrudin, line 10: This 3rd method a really nice addition, and I think it means the following things:

1. The CCH implementation can be given the currentQuery and can modify it AND can hold onto it as a private field
2. injectConstraintIntoQuery doesn't need to be exposed in this interface
3. getMaxConstraintValue doesn't need currentQuery passed to it because the CCH already knows what it is
4. (You're not going to like this part) I don't think we need a QueryContext anymore :) . That's because if CCH is holding onto the currentQuery, it doesn't need to go into QueryContext, which means we don't need QueryContext anymore, we only need a RowBatcher and a RowBatcherBuilder. 

I think that means that what CCH is really about is managing the user's query... though it can't be called QueryManager since the Java Client already uses that term. But based on your PR, it fulfills two needs for the Task class:

1. It takes the user's query and makes the correct call to withBatchView on the RowBatcher (including support for the constraintColumn)
2. If the user specified a constraint column, the Task class will call getMaxConstraintValue

I think `QueryHandler` is what we want - it basically subsumes QueryContext and ConstraintColumnHandler. Ping me for more chat about this, but I think that's where the solution is evolving towards:

```
interface QueryHandler
  String addQueryToRowBatcher(String userQuery, RowBatcher)
  String getMaxConstraintValue(long queryStartTimeInMillis, String constraintColumnName)
```

Note that I think we can safely use the name "userQuery" in the interface, though the QH impl will likely call this "currentQuery" in its implementation since it may have been modified. And I think `addQuery` can return the query to make it easier to test this.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -57,12 +57,15 @@ public final void start(Map<String, String> config) {
         this.databaseClient = new DefaultConfiguredDatabaseClientFactory().newDatabaseClient(databaseClientConfig);
         dataMovementManager = databaseClient.newDataMovementManager();
         pollDelayMs = (Long) parsedConfig.get(MarkLogicSourceConfig.WAIT_TIME);
-        this.constraintColumn = (String) parsedConfig.get(MarkLogicSourceConfig.CONSTRAINT_COLUMN_NAME);
+        constraintColumn = (String) parsedConfig.get(MarkLogicSourceConfig.CONSTRAINT_COLUMN_NAME);
         if (!StringUtils.hasText(constraintColumn)) {
             constraintColumn = null;
+        } else {
+            constraintColumn = constraintColumn.replaceAll("[\"']", "");
```
rjrudin, line 64: This line of code is repeated in DslConstraintColumnHandler, which means it's a good candidate to be stuffed into a public static "util" method somewhere where it can also be easily unit-tested to cover what it's intended to handle.

src/main/java/com/marklogic/kafka/connect/source/SerializedConstraintColumnHandler.java
```diff
@@ -0,0 +1,92 @@
...
+            try {
+                ObjectNode constraintNode = (ObjectNode) mapper.readTree(constraintNodeString);
+
+                ObjectNode queryRoot = (ObjectNode) mapper.readTree(originalQuery);
+                ArrayNode rootArgsArray = (ArrayNode) queryRoot.get("$optic").get("args");
+
+                rootArgsArray.insert(1, constraintNode);
+                constrainedQuery = mapper.writeValueAsString(queryRoot);
+            } catch (JsonProcessingException e) {
+                throw new RuntimeException(e);
```
rjrudin, line 41: We usually want friendly contextual messages, so e.g. - `throw new RuntimeException("Unable to modify serialized query, cause: " + e.getMessage(), e)`

src/main/java/com/marklogic/kafka/connect/source/SerializedConstraintColumnHandler.java
```diff
@@ -0,0 +1,92 @@
...
+        RawPlanDefinition maxConstraintValueQuery = rowMgr.newRawPlanDefinition(new StringHandle(currentMaxValueQuery));
+        JacksonHandle handle = new JacksonHandle().withFormat(Format.JSON).withMimetype("application/json");
+        handle.setPointInTimeQueryTimestamp(queryStartTimeInMillis);
+        JacksonHandle result = rowMgr.resultDoc(maxConstraintValueQuery, handle);
+        String newMaxConstraintValue = "";
+        try {
+            newMaxConstraintValue = result.get().get("rows").get(0).get("constraint").get("value").asText();
+            newMaxConstraintValue = newMaxConstraintValue.replaceAll("[\"']", "");
+        } catch (Exception ex) {
+            logger.warn("Failed to get a valid Maximum Constraint value: " + ex.getMessage());
```
rjrudin, line 62: This is an interesting error scenario. If we fail to get the max value but we don't throw an error and instead return N rows, then on the next poll(), those same N rows will be returned. I am thinking the default should be to throw an error (which will be caught by `poll()`) so that no rows are returned, but I'm not 100% sure about that. It may need to be a new config option, which would be a new story - i.e. whether the user wants an error to be thrown or not in this scenario.

src/main/java/com/marklogic/kafka/connect/source/SerializedConstraintColumnHandler.java
```diff
@@ -0,0 +1,92 @@
...
+            ObjectNode limitOneNode = (ObjectNode) mapper.readTree(limitOnePhase);
+            String constraintColumnPhase = "{\"ns\":\"op\", \"fn\":\"select\", \"args\":[[{\"ns\":\"op\", \"fn\":\"as\", \"args\":[\"constraint\", {\"ns\":\"op\", \"fn\":\"col\", \"args\":[\"" + constraintColumn + "\"]}]}]]}";
+            ObjectNode constraintColumnNode = (ObjectNode) mapper.readTree(constraintColumnPhase);
+            ObjectNode queryRoot = (ObjectNode) mapper.readTree(currentQuery);
+            ArrayNode rootArgsArray = (ArrayNode) queryRoot.get("$optic").get("args");
+            rootArgsArray.add(orderByDescendingNode);
+            rootArgsArray.add(limitOneNode);
+            rootArgsArray.add(constraintColumnNode);
+            return mapper.writeValueAsString(queryRoot);
+        } catch (JsonProcessingException e) {
```
rjrudin, line 81: I think this method should declare `throws JsonProcessingException` and then the one calling method should include it in its try/catch so that any error involved in trying to get the max value is handled the same way (right now, this would be handled differently because it will propagate up to poll() and result in no rows being returned).

src/test/java/com/marklogic/kafka/connect/source/SerializedConstraintInjectionTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+class SerializedConstraintInjectionTest extends AbstractIntegrationSourceTest {
+    private static final String constraintColumn = "lucky_number";
+    private static final String constraintValue = "52";
+
+    private final SerializedConstraintColumnHandler serializedConstraintColumnHandler = new SerializedConstraintColumnHandler(null);
+
+    @Test
+    void testAccessorOnlyQuery() {
+        String originalQuery = "{\"$optic\":{\"ns\":\"op\", \"fn\":\"operators\", \"args\":[{\"ns\":\"op\", \"fn\":\"from-view\", \"args\":[\"Medical\", \"Authors\"]}]}}";
```
rjrudin, line 17: These strings of JSON are pretty difficult to read and understand. It would help if they're pretty-printed... and I think that will be true in production too - i.e. if we ever want to log something that includes the serialized plan, it'll be far easier to understand if it's pretty-printed. 

I think each of these tests should do what the test below this one does - pretty-print the input and toss it into a String (or what may be better is to toss it into a file under src/test/resources) so that a reader of the test can easily understand what's in the original plan and what's in the expected plan. And then the Serialized implementation should be changed to call "toPrettyString" on the ObjectNode it builds up. There's a very small cost to doing toPrettyString instead of toString that will be dwarfed by the amount of time spent actually getting data back. 

It may be good to have one test where the input is a single line of JSON, which will likely be the case when using a properties file, just to ensure that works. But the expected output would still be pretty-printed.

src/test/java/com/marklogic/kafka/connect/source/SerializedConstraintInjectionPollTest.java
```diff
@@ -0,0 +1,158 @@
...
+            "The max value after polling once should be '2022-04-11T11:00:00'");
+
+        task.poll();
+        previousMaxConstraintColumnValue = task.getPreviousMaxConstraintColumnValue();
+        Assertions.assertEquals("2022-05-11T10:00:00", previousMaxConstraintColumnValue,
+            "The max value after polling twice should be '2022-05-11T10:00:00'");
+    }
+
+    @Test
+    void testSecondPollWithNewRowsOnlyReturnsRowAfterPreviousBatchMaxValue() throws IOException, InterruptedException {
```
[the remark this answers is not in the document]
BillFarber: Essential Enterprise 11.0.0

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -141,19 +145,22 @@ private boolean rowBatcherErrorIsKnownServerBug(Exception ex) {
...
-    private String getBatchMaxValue(String currentQuery, long queryStartTimeInMillis, String constraintColumn) {
-        String currentMaxValueQuery = buildMaxValueDslQuery(currentQuery, constraintColumn);
-
-        RowManager rowMgr = databaseClient.newRowManager();
-        RawQueryDSLPlan plan = rowMgr.newRawQueryDSLPlan(new StringHandle(currentMaxValueQuery));
-        JacksonHandle handle = new JacksonHandle().withFormat(Format.JSON).withMimetype("application/json");
-        handle.setPointInTimeQueryTimestamp(queryStartTimeInMillis);
-        JacksonHandle result = rowMgr.resultDoc(plan, handle);
-        return result.get().get("rows").get(0).get("constraint").get("value").toString();
+    public ConstraintColumnHandler newConstraintColumnHandler(Map<String, Object> parsedConfig) {
```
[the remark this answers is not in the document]
BillFarber: I didn't know you could do this.
Done.

src/main/java/com/marklogic/kafka/connect/source/SerializedConstraintColumnHandler.java
BillFarber: Since 1) it's conceivable that this could fail, but be recoverable (network issue) and 2) the customer could be okay with duplicate records,
Then, I lean toward defaulting to a log message (as above), but creating a config value.

src/main/java/com/marklogic/kafka/connect/source/DslQueryHandler.java
```diff
@@ -0,0 +1,84 @@
...
+import com.marklogic.client.io.Format;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.io.StringHandle;
+import com.marklogic.client.row.RawQueryDSLPlan;
+import com.marklogic.client.row.RowManager;
+import org.springframework.util.StringUtils;
+
+import java.util.Map;
+
+public class DslQueryHandler extends LoggingObject implements QueryHandler {
```
rjrudin, line 15: I'll look at this more closely next week, but one thing I like right away is that there aren't any Kafka dependencies in here (the config map is related to Kafka, but that could easily be broken out into the actual params that the handler needs). I think that's important because this capability - modifying a user's query and then getting the highest value for a particular column - is something we may want to migrate into the Java Client itself. For example, the similar logic for doing that in the NiFi connector with the Search API could arguably in the Java Client too.

src/main/java/com/marklogic/kafka/connect/source/QueryHandler.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static java.lang.String.format;
+
+public interface QueryHandler {
+    void addQueryToRowBatcher(RowBatcher<?> rowBatcher);
+
+    String updatePreviousMaxConstraintColumnValue(long queryStartTimeInMillis);
+
+    String ANY_QUOTES_REGEX_PATTERN = "[\"']";
+    default String sanitize(String target) {
```
rjrudin, line 18: The sanitize and removeWhitespace methods can live here for now, but because they're not part of the public interface, they should be elsewhere, like in a non-public QueryHandlerUtil class. The newQueryHandler method is public though as it's the expected factory method for obtaining a QueryHandler.

src/main/java/com/marklogic/kafka/connect/source/QueryHandler.java
```diff
@@ -0,0 +1,44 @@
...
+        boolean configuredForDsl = StringUtils.hasText((String) parsedConfig.get(MarkLogicSourceConfig.DSL_QUERY));
+        boolean configuredForSerialized = StringUtils.hasText((String) parsedConfig.get(MarkLogicSourceConfig.SERIALIZED_QUERY));
+        QueryHandler queryHandler;
+        if ((!(configuredForDsl || configuredForSerialized)) || (configuredForDsl && configuredForSerialized)) {
+            throw new ConfigException(
+                format("Either a DSL Optic query (%s) or a serialized Optic query (%s), but not both, are required",
+                    MarkLogicSourceConfig.DSL_QUERY, MarkLogicSourceConfig.SERIALIZED_QUERY)
+            );
+        }
+        if (configuredForDsl) {
```
rjrudin, line 37: I think this is where the ternary operator can cut down on some Java verbosity - i.e. you can avoid the queryHandler declaration above and just do :

```
return configuredForDsl ? 
    new Dsl... : 
    new Serialized ...
```

I almost always go for that instead of declaring a variable for the sole purpose of returning it.

src/main/java/com/marklogic/kafka/connect/source/CsvQueryContextBuilder.java
```diff
@@ -14,27 +14,21 @@
 import java.util.List;
 import java.util.Map;
 
-public class CsvQueryContextBuilder extends AbstractRowBatchBuilder implements QueryContextBuilder<String> {
+public class CsvQueryContextBuilder extends AbstractRowBatchBuilder <String> {
```
rjrudin, line 17: Not urgent - can be changed in a subsequent PR - but the subclasses should go back to "(Type)RowBatcherBuilder" and the parent class needs to have "Batch" changed to "Batcher".

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -54,63 +48,54 @@ public final void start(Map<String, String> config) {
...
+        queryHandler = QueryHandler.newQueryHandler(databaseClient, parsedConfig);
+        rowBatcherBuilder = AbstractRowBatchBuilder.newRowBatcherBuilder(dataMovementManager, parsedConfig);
         logger.info("Started RowBatcherSourceTask");
     }
 
     @Override
     public List<SourceRecord> poll() throws InterruptedException {
         // Temporary logging while testing CP to ensure the correct output format is being used
-        logger.info("Polling; QueryContextBuilder: " + queryContextBuilder.getClass().getName());
+        logger.info("Polling ; RowBatcherBuilder: " + rowBatcherBuilder.getClass().getName());
```
rjrudin, line 63: Just do one log event here; I keep thinking to my previous job with using Loki + Grafana to look at log events, and having one logical log event spread across multiple lines was a real pain. There's only one log event here - logging the fact that we're polling along with a couple details - so we only want one log message.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -54,63 +48,54 @@ public final void start(Map<String, String> config) {
...
+        databaseClient = new DefaultConfiguredDatabaseClientFactory().newDatabaseClient(databaseClientConfig);
         dataMovementManager = databaseClient.newDataMovementManager();
         pollDelayMs = (Long) parsedConfig.get(MarkLogicSourceConfig.WAIT_TIME);
-        this.constraintColumn = (String) parsedConfig.get(MarkLogicSourceConfig.CONSTRAINT_COLUMN_NAME);
-        if (!StringUtils.hasText(constraintColumn)) {
-            constraintColumn = null;
-        }
 
-        queryContextBuilder = AbstractRowBatchBuilder.newQueryContextBuilder(dataMovementManager, parsedConfig);
+        queryHandler = QueryHandler.newQueryHandler(databaseClient, parsedConfig);
```
rjrudin, line 55: I don't think this should be initialized on `start`, but rather it should be created each time (like RBB) in `poll`. I think the flow should be in `poll`:

1. Call `QueryHandler.newQueryHandler` to obtain a new QH (which will only be a local variable in `poll`)
2. Call `addQueryToRowBatcher`, passing in the "previous max value" which the Task is responsible for obtaining access to (since it will eventually live in ML and be retrieved from there on every `poll` call) . As part of this call, the QH modifies the user's query to include the constraint column and previous max value (both of which may be null, in which case it does nothing).
3. Task runs the job
4. Task then calls a `getMaxConstraintColumnValue` function on QH that uses the modified user query to get the max value, and the Task stores that 

The QH thus lives only for the duration of the `poll()` method and thus should not be a class attribute here. And while the QH is responsible for getting the max value, the Task is responsible for storing it - currently as a Task class attribute, eventually and optionally as a value in ML.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -54,63 +48,54 @@ public final void start(Map<String, String> config) {
...
             dataMovementManager.startJob(rowBatcher);
             logger.info("Awaiting completion");
             rowBatcher.awaitCompletion();
             dataMovementManager.stopJob(rowBatcher);
 
-            if (constraintColumn != null) {
-                long queryStartTimeInMillis = rowBatcher.getServerTimestamp();
-                previousMaxConstraintColumnValue = getBatchMaxValue(queryContext.getCurrentQuery(), queryStartTimeInMillis, constraintColumn);
-            }
+            long queryStartTimeInMillis = rowBatcher.getServerTimestamp();
```
rjrudin, line 93: The Task knows it only needs to do this if a constraint column option has been set, so I think it should do an `if` statement here to only call this stuff if that option has been set.

src/test/java/com/marklogic/kafka/connect/RegexTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.marklogic.kafka.connect.source.DslQueryHandler;
+import com.marklogic.kafka.connect.source.MarkLogicSourceConfig;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class RegexTest {
```
rjrudin, line 11: I think this name is a good indicator that those methods in the interface above should really be in an e.g. `QueryHandlerUtil` non-public class, and this then becomes `QueryHandlerUtilTest` which is a more self-documenting name than `RegexTest`, and should live in the "source" package as well.

rjrudin (changes requested):

rjrudin (approved):
Looks good to merge, I'll do the QA on this tomorrow

PR #108, opened 2022-12-30
DEVEXP-131: Configure constraining column value storage in MarkLogic

BillFarber:
DEVEXP-131: Configure constraining column value storage in MarkLogic

### DIFF
--- a/src/main/java/com/marklogic/kafka/connect/source/MarkLogicConstraintValueStore.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/MarkLogicConstraintValueStore.java
@@ -1,0 +1,83 @@
+package com.marklogic.kafka.connect.source;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.FailedRequestException;
+import com.marklogic.client.document.DocumentDescriptor;
+import com.marklogic.client.document.JSONDocumentManager;
+import com.marklogic.client.ext.util.DefaultDocumentPermissionsParser;
+import com.marklogic.client.io.DocumentMetadataHandle;
+import com.marklogic.client.io.Format;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.io.StringHandle;
+import org.springframework.util.StringUtils;
+
+import java.util.Map;
+
+public class MarkLogicConstraintValueStore extends ConstraintValueStore {
+    private final DefaultDocumentPermissionsParser permissionsParser = new DefaultDocumentPermissionsParser();
+    private final DatabaseClient databaseClient;
+    private final String constraintStorageUri;
+    private String constraintStoragePermissions;
+    private String constraintStorageCollections;
+
+    MarkLogicConstraintValueStore(DatabaseClient databaseClient, String constraintStorageUri, String constraintColumn,
+                                  Map<String, Object> parsedConfig) {
+        super(constraintColumn);
+        this.databaseClient = databaseClient;
+        this.constraintStorageUri = constraintStorageUri;
+        constraintStoragePermissions = (String) parsedConfig.get(MarkLogicSourceConfig.CONSTRAINT_STORAGE_PERMISSIONS);
+        if (!StringUtils.hasText(constraintStoragePermissions)) {
+            constraintStoragePermissions = null;
+        }
+        constraintStorageCollections = (String) parsedConfig.get(MarkLogicSourceConfig.CONSTRAINT_STORAGE_COLLECTIONS);
+        if (!StringUtils.hasText(constraintStorageCollections)) {
+            constraintStorageCollections = null;
+        }
+    }
+
+    public void storeConstraintState(String previousMaxConstraintColumnValue, int lastRowCount) {
+        String constraintStateJson = "";
+        try {
+            constraintStateJson = buildConstraintState(previousMaxConstraintColumnValue, lastRowCount);
+            StringHandle handle = new StringHandle(constraintStateJson).withFormat(Format.JSON);
+            DocumentMetadataHandle metadataHandle = new DocumentMetadataHandle();
+            insertPermissions(metadataHandle);
+            insertCollections(metadataHandle.getCollections());
+            JSONDocumentManager mgr = databaseClient.newJSONDocumentManager();
+            mgr.write(constraintStorageUri, metadataHandle, handle);
+        } catch (JsonProcessingException e) {
+            String errorMessage = "Unable to store constraint value (" + constraintStateJson + ") at URI: "
+                + constraintStorageUri + "; cause: " + e.getMessage();
+            throw new RuntimeException(errorMessage, e);
+        } catch (FailedRequestException e) {
+            String errorMessage = "Unable to store constraint value (" + constraintStateJson + ") at URI: "
+                + constraintStorageUri + "; cause: " + e.getMessage();
+            throw new RuntimeException(errorMessage, e);
+        }
+    }
+
+    private void insertPermissions(DocumentMetadataHandle metadataHandle) {
+        if (StringUtils.hasText(constraintStoragePermissions)) {
+            permissionsParser.parsePermissions(constraintStoragePermissions, metadataHandle.getPermissions());
+        }
+    }
+
+    private void insertCollections(DocumentMetadataHandle.DocumentCollections documentCollections) {
+        if (StringUtils.hasText(constraintStorageCollections)) {
+            documentCollections.addAll(constraintStorageCollections.split(","));
+        }
+    }
+
+    public String retrievePreviousMaxConstraintColumnValue() {
+        JacksonHandle resultHandle = new JacksonHandle();
+        DocumentDescriptor descriptor = databaseClient.newJSONDocumentManager().exists(constraintStorageUri);
+        if (descriptor != null) {
+            databaseClient.newJSONDocumentManager().read(constraintStorageUri, resultHandle);
+            return resultHandle.get().findPath("marklogicKafkaConstraintLastValue").textValue();
+        } else {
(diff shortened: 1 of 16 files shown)


src/test/java/com/marklogic/kafka/connect/source/ConstraintStorageTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import com.marklogic.junit5.PermissionsTester;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.List;
+
+class ConstraintStorageTest extends AbstractIntegrationSourceTest {
```
rjrudin, line 11: I think these two tests can be merged, as there's really just one scenario:

1. Load the fifteen authors
2. Run the Task
3. Verify that the constraint value document was written correctly (as you're doing already)
4. Load 3 more authors - two with IDs higher than the highest one that was captured, and one with an ID that is lowest that the highest one
5. Run the Task again
6. Verify that only the two new authors with a higher ID are returned
7. And then verify that the constraint value doc was updated as well

rjrudin: Also, I think it's worth adding an error scenario too - i.e. do something to cause the read of the JSON doc to fail, such as first writing some random XML to that URI. Then load the 15 authors, and run the Task, and verify that you get null back since the read of the JSON doc failed.

rjrudin (changes requested):
This looks good, and while I haven't QA'ed it yet, looks functionally correct.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicConstraintValueStore.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    public void storeConstraintState(String previousMaxConstraintColumnValue, int lastRowCount) {
+        String constraintStateJson = "";
+        try {
+            constraintStateJson = buildConstraintState(previousMaxConstraintColumnValue, lastRowCount);
+            JSONDocumentManager JSONDocMgr = databaseClient.newJSONDocumentManager();
+            StringHandle handle = new StringHandle(constraintStateJson);
+            DocumentMetadataHandle metadataHandle = new DocumentMetadataHandle();
```
rjrudin, line 43: This use case of - build up a doc to be written to ML, based on some String properties - seems like another that the Java Client should support. We actually have a class to do this already in the Kafka project - `DocumentWriteOperationBuilder` - which I think you can reuse here, but don't worry about it in this PR - it's a refactor for later, and I want to think more about whether the Java Client should really be providing this functionality because we know it's a common use case (e.g. the NiFi connector does this exact same thing too).

rjrudin (commented):
This looks good, I'll do some quick QA on it this morning and approve it after that.

src/test/java/com/marklogic/kafka/connect/source/ConstraintStorageTest.java
```diff
@@ -0,0 +1,87 @@
...
+        loadSingleAuthorRowIntoMarkLogicWithCustomTime("thirdHigherId", "8", evenLaterTime, "thirdHigherId");
+        loadSingleAuthorRowIntoMarkLogicWithCustomTime("fourthHigherId", "9", evenLaterTime, "fourthHigherId");
+
+        storeBadConstraintState(constraintStorageUri);
+        String badLastValue = constraintValueStore.retrievePreviousMaxConstraintColumnValue();
+        Assertions.assertNull(badLastValue, "The retrieved last value should be null since the document is invalid");
+
+        newRecords = task.poll();
+        Assertions.assertEquals(20, newRecords.size(),
+            "Since the stored state document is invalid, all 21 records should have been returned");
```
rjrudin, line 78: Typo, should be "20" - though I would just do "all" to avoid the chance of the number being incorrect

src/test/java/com/marklogic/kafka/connect/source/ConstraintStorageTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void storeBadConstraintState(String constraintStorageUri) {
```
rjrudin, line 81: This should be private, just to make it clear it's not used anywhere else. Also, this is going to fail not because the data is invalid, but because the doc is lacking a read permission and the user won't see it. I think it'd be better, and less code, to just reuse the ConstraintValueStore but pass in a value that isn't valid. 

Actually - I just tried this, and it appears to be a different scenario. I tried this:

    constraintValueStore.storeConstraintState("invalid-value");

And the behavior I got instead was that no rows were returned. 

That seems unexpected to me - i.e. I would expect that if there's any problem with the constraint value document (it's missing, it's invalid, or the doc is valid JSON but the value is invalid), I would get the same behavior - which I think should be that all rows are returned. 

I think that's a good scenario to add, and I think both that and this scenario of "Doc is invalid" should be in separate test methods. It makes this method easier to understand (it's 60 LOC, so a little long already), and makes it easier to understand what happens for these two error scenarios. 

A 4th error scenario - try writing the doc with no permissions. That should fail, and we want to document what happens there.

src/test/java/com/marklogic/kafka/connect/source/ConstraintStorageTest.java
```diff
@@ -0,0 +1,87 @@
...
+import com.marklogic.junit5.PermissionsTester;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+
+class ConstraintStorageTest extends AbstractIntegrationSourceTest {
```
rjrudin, line 14: To be consistent with story/action-based test names, let's do "StoreConstraintValueInMarkLogicTest".

rjrudin (commented):
Few comments, doing QA in CP now.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -19,6 +22,9 @@ public class MarkLogicSourceConfig extends MarkLogicConfig {
     public static final String DSL_QUERY = "ml.source.optic.dsl";
     public static final String SERIALIZED_QUERY = "ml.source.optic.serialized";
     public static final String CONSTRAINT_COLUMN_NAME = "ml.source.optic.constraintColumn.name";
+    public static final String CONSTRAINT_STORAGE_URI = "ml.source.optic.constrainingColumn.uri";
```
rjrudin, line 25: This is correct per the story, but the story is now a little out-of-date (no need to update it) since we landed on "constraintColumn" above. So for consistency, these should all be "constraintColumn" too.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicConstraintValueStore.java
```diff
@@ -0,0 +1,82 @@
...
+            constraintStorageCollections = null;
+        }
+    }
+
+    public void storeConstraintState(String previousMaxConstraintColumnValue, int lastRowCount) {
+        String constraintStateJson = "";
+        try {
+            constraintStateJson = buildConstraintState(previousMaxConstraintColumnValue, lastRowCount);
+            JSONDocumentManager JSONDocMgr = databaseClient.newJSONDocumentManager();
+            StringHandle handle = new StringHandle(constraintStateJson);
```
rjrudin, line 42: This should use `.withFormat(Format.JSON)` so that ML knows to create a JSON document as opposed to relying on the user-defined URI.

src/main/java/com/marklogic/kafka/connect/source/RowBatcherSourceTask.java
```diff
@@ -90,10 +93,11 @@ protected void performPoll(QueryHandler queryHandler, List<SourceRecord> newSour
...
             dataMovementManager.stopJob(rowBatcher);
 
-            if (constraintColumn != null && !newSourceRecords.isEmpty()) {
+            if (constraintValueStore != null && !newSourceRecords.isEmpty()) {
                 long queryStartTimeInMillis = rowBatcher.getServerTimestamp();
-                previousMaxConstraintColumnValue = queryHandler.updatePreviousMaxConstraintColumnValue(queryStartTimeInMillis);
-                logger.info("Storing new max constraint value: " + previousMaxConstraintColumnValue);
+                String newMaxConstraintColumnValue = queryHandler.updatePreviousMaxConstraintColumnValue(queryStartTimeInMillis);
+                logger.info("Storing new max constraint value: " + newMaxConstraintColumnValue);
+                constraintValueStore.storeConstraintState(newMaxConstraintColumnValue, newSourceRecords.size());
```
rjrudin, line 100: I did a test here where I forced a failure by setting the uri property to "hi.xml" - ML complains because it assumes I want an XML document. But it could fail for other reasons too, like missing an update permissions.

The error then gets logged, stating "job failed to complete successfully" - but the source records that were retrieved are still returned. That would be surprising to a user - if the job failed (though it didn't, the job succeeded), I think a user would expect no data to be returned. 

I think that suggests the following changes:

1. The try/catch should only be around this block of code for getting and storing the max value
2. The error message should be specific to that goal and should be a warning instead of an error since we're still going to return data - i.e. "Unable to update max constraint value, cause: " + e.getMessage()

rjrudin (changes requested):

src/test/java/com/marklogic/kafka/connect/source/ConstraintStorageTest.java
BillFarber: I'm not real sure I've covered what you're looking for. Maybe we should take a look at it together.

rjrudin (approved):

PR #112, opened 2023-01-05
Switching from RowBatcher to RowManager

rjrudin:
Consensus was that the downsides of RowBatcher - can only use fromView; errors happening in the background; confusion over what "batchSize" refers to - were not worth the potential benefit in performance. That's because once RowBatcher is noticeably faster than RowManager, the user is likely pulling back enough rows to run out of memory / Java heap space due to a massive List. 

This leaves all the RowBatcher support in place for now. Main addition is a `PlanInvoker` that does what `RowBatcherBuilder` effectively did, which is to choose the correct handle, invoke a plan, and convert the results into source records. 

Once this is merged, I'll submit a PR to remove all the RB support and rename the task to RowManagerSourceTask. Having that all in one commit/PR will make it easy to pull from if we decide to add RB support back in the future.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ pipeline{
   stages{
     stage('tests'){
       steps{
-        copyRPM 'Release','10.0-9.4'
+        copyRPM 'Latest','11.0'
         setUpML '$WORKSPACE/xdmp/src/Mark*.rpm'
         sh label:'setup', script: '''#!/bin/bash
         cd kafka-connector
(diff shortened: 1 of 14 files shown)


src/main/java/com/marklogic/kafka/connect/source/CsvPlanExecutor.java
```diff
@@ -0,0 +1,40 @@
...
+import com.marklogic.client.io.StringHandle;
+import org.apache.kafka.connect.source.SourceRecord;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CsvPlanExecutor implements PlanExecutor {
```
rjrudin, line 15: I made this new PlanExecutor interface and copied the code of interest from each RBB to the appropriate PE impl.

BillFarber (approved):
LGTM

PR #115, opened 2023-01-09
DEVEXP-139: View query in logged error when poll fails

BillFarber:
I think this is all we need.

### DIFF
--- a/src/main/java/com/marklogic/kafka/connect/source/RowManagerSourceTask.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/RowManagerSourceTask.java
@@ -64,12 +64,14 @@ public class RowManagerSourceTask extends SourceTask {
         logger.info("Polling; sleep time: {}ms", pollDelayMs);
         Thread.sleep(pollDelayMs);
 
+        String currentQuery = "<Not Built Yet>";
         try {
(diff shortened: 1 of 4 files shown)


src/main/java/com/marklogic/kafka/connect/source/RowManagerSourceTask.java
```diff
@@ -78,8 +80,8 @@ public List<SourceRecord> poll() throws InterruptedException {
             updateMaxConstraintValue(results, queryHandler);
             return newSourceRecords.isEmpty() ? null : newSourceRecords;
         } catch (Exception ex) {
-            // TODO Will include query here soon
-            logger.error("Unable to poll for source records; cause: " + ex.getMessage(), ex);
+            logger.error("Unable to poll for source records; cause: " + ex.getMessage() +
+                "\nCurrent Query: <" + currentQuery + ">", ex);
```
rjrudin, line 84: Remove the newline symbol, as that can give log analyzers issues too - i.e. they might see the newline as a new log event and not as a continuation of the current log event. Remove the "<" and ">" symbols too, as we don't use markers like that in any other log message - just log whatever the query is, that'll suffice.

rjrudin (changes requested):

PR #116, opened 2023-01-09
Feature/document source with kafka

BillFarber:
Just a draft to verify that I'm going in the right direction with this documentation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -259,15 +259,21 @@ and
 
     bin/kafka-server-start.sh config/server.properties
 
-Next, start the Kafka connector in standalone mode (also from the Kafka home directory):
+Next, start the Kafka connector in standalone mode (also from the Kafka home directory). To run the sink connector, 
+use the following command:
 
     bin/connect-standalone.sh config/marklogic-connect-standalone.properties config/marklogic-sink.properties
 
-You'll see a fair amount of logging from Kafka itself; near the end of the logging, look for messages from
-`MarkLogicSinkTask` and MarkLogic Java Client classes such as `WriteBatcherImpl` to ensure that the connector has
-started up correctly.
+To run the source connector, the command is:
 
-To test out the connector, you can use the following command to enter a CLI that allows you to manually send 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -280,3 +286,15 @@ When a document is received and written by the connector, you'll see logging lik
...
+
+## Source Connector Testing
+To test out the source connector, you can insert documents into MarkLogic that generate Optic rows that match your query.
+Then, after the connector is running, monitor the target.
+
+    TODO - Show example Optic DSL plan
+
+When a document is received and written by the connector, you'll see logging like this:
+
+    TODO - Do we want a link to TDE & Optic?
```
rjrudin, line 298: I think we have plenty of links for those already in the README

rjrudin (commented):
Looks good so far!

CONTRIBUTING.md
```diff
@@ -280,3 +286,17 @@ When a document is received and written by the connector, you'll see logging lik
...
+
+## Source Connector Testing
+To test out the source connector, you can insert documents into MarkLogic that contain data that matches TDE templates
+to generate Optic rows that match your query. For example, if your TDE template defines an "Authors" view in the
+"Medical" schema, the source configuration file might have the following properties.
+
+    ml.source.optic.dsl=op.fromView("Medical", "Authors")
+    ml.source.topic=Authors
+
+Then, after the connector is running, monitor the "Authors" topic. When a document is retrieved from MarkLogic and
```
rjrudin, line 298: How does the user actually monitor the topic? Is there an equivalent process to bin/kafka-console-producer.sh that the user can use to see what's in a topic? I don't think the logging will suffice because the user primarily wants to see what the data looks like that ends up in a topic. A CP user can do that view the "Messages" tab for a topic. 

I just saw there's a kafka-console-consumer.sh tool - that sounds like it would do the job and is what we'd then want to mention here.

rjrudin (changes requested):

rjrudin (approved):

PR #119, opened 2023-01-11
DEVEXP-160: Document how DLQ support works in the connector

BillFarber:


README.md
```diff
@@ -84,6 +84,25 @@ two properties in the "Common" section
 copy of the `./config/marklogic-connect-standalone.properties` file. This file already has the two properties set to the
   correct values, so you should not need to do anything further with them.
 
+### Dead Letter Configuration
+
+The connector may also be configured to use a dead letter queue (DLQ) when the connector is unable to successfully
```
rjrudin, line 89: I think we want a different approach here - this is explaining how to setup DLQ, and we don't want to be in the business of explaining how to configure Kafka - the user is either assumed to know how to do that, or is expected to figure that out from Kafka/Confluent docs. At the most, providing the link to https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues/ is enough .

Instead, what we want to tell the user is that the sink connector will now (as of 1.8.0) make use of the DLQ if it's been configured. And we have two events to identify:

1. When an incoming Kafka record cannot be converted into a document to write to MarkLogic
2. When a batch of documents (converted from Kafka records) fails to be written to MarkLogic

For 1, we need to explain that only the one failed record will be sent to the DLQ.

For 2, we need to explain that the entire batch of records will be sent to the DLQ, as the connector is not aware of which document caused the failure.

rjrudin (changes requested):

README.md
rjrudin: We should also document the headers that we add to a record sent to the DLQ, that would be of interest to the user.

README.md
```diff
@@ -84,6 +84,21 @@ two properties in the "Common" section
 copy of the `./config/marklogic-connect-standalone.properties` file. This file already has the two properties set to the
   correct values, so you should not need to do anything further with them.
 
+### Dead Letter Configuration
+
+Starting with version 1.8.0, the sink connector makes use of the dead letter queue (DLQ) if the user has configured
+Kafka to use the DLQ. For information about configuring Kafka to use the DLQ, please refer to:
+
+    https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues/
```
rjrudin, line 92: Putting this in a code block makes it unclickable - see:

image

I recommend doing this instead (i.e. helpful text in brackets, followed by link in parens):

[Please see the Kafka documentation](https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues/) for more information on configuring the dead letter queue in Kafka.

README.md
```diff
@@ -84,6 +84,21 @@ two properties in the "Common" section
...
+
+When Kafka has been configured to use the DLQ, there are two events in the sink connector that will cause a record to be
+sent to the DLQ.
+- A conversion problem: If a specific record cannot be converted into the target format to be delivered to MarkLogic,
+then that record will be sent to the DLQ.
+- A write problem: If a batch of documents (converted Kafka records from the source topic) fails to be written to
+MarkLogic then each of the records in the batch will be sent to the DLQ. The entire batch must be sent to the DLQ since
+the connector is unable to determine the cause of the failure.
```
rjrudin, line 100: Is it not worth documenting the headers that are added? Those seem worth documenting to me, as it's our attempt to provide some assistance with trying to figure out why the record failed.

BillFarber: Sorry, forgot this one.

README.md
```diff
@@ -84,6 +84,27 @@ two properties in the "Common" section
...
+
+When Kafka has been configured to use the DLQ, there are two events in the sink connector that will cause a record to be
+sent to the DLQ.
+- "Record conversion" : If a specific record cannot be converted into the target format to be delivered to MarkLogic,
+then that record will be sent to the DLQ.
+- "Write Failure" : If a batch of documents (converted Kafka records from the source topic) fails to be written to
+MarkLogic then each of the records in the batch will be sent to the DLQ. The entire batch must be sent to the DLQ since
+the connector is unable to determine the cause of the failure.
+
+When a record is sent to the DLQ, the connector first adds headers to the record provide information about the cause of
```
rjrudin, line 101: provide -> providing

README.md
```diff
@@ -84,6 +84,27 @@ two properties in the "Common" section
...
+the failure in order to assist with troubleshooting and potential routing.
+- "marklogic-failure-type" : Either "Write Failure" or "Record conversion"
```
rjrudin, line 103: I think the code should be adjusted so that consistent capitalization is used here - I think "Record conversion" is better, so change the code and docs to "Write failure"

rjrudin (changes requested):

rjrudin (approved):

PR #120, opened 2023-01-11
DEVEXP-161: Update connector manifest.json file

BillFarber:
DEVEXP-161: Update connector manifest.json file

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,6 +1,7 @@
 {
   "component_types": [
-    "sink"
+    "sink",
+    "source"
   ],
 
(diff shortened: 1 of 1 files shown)


rjrudin (approved):

manifest.json
```diff
@@ -36,13 +37,13 @@
   },
 
   "requirements": [
-    "MarkLogic 9+"
+    "MarkLogic 10+"
```
rjrudin, line 40: Note that the sink connector should still be fine on ML 9 (and likely ML 8 too); an array is supported here, so maybe it's better to do "MarkLogic 9 for sink", "MarkLogic 10 for source" or something like that?

rjrudin (approved):

PR #121, opened 2023-01-12
Accessor tests

BillFarber:
New tests to start covering all the Optic data access functions.

Take particular notice of "pullDataUsingFromLexicons". This one doesn't just verify the injection, but also uses the injection to retrieve limited data.

Also take notice of BALANCED_PARENS_REGEX_PATTERN and it's use in DslQueryHandler.injectConstraintIntoQuery. I believe it gives me the first op function (no matter what it contains).

### DIFF
--- a/src/main/java/com/marklogic/kafka/connect/source/DslQueryHandler.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/DslQueryHandler.java
@@ -12,6 +12,7 @@ import org.springframework.util.StringUtils;
 
 import java.util.Map;
 
+
 public class DslQueryHandler extends LoggingObject implements QueryHandler {
 
     private final DatabaseClient databaseClient;
@@ -49,9 +50,12 @@ public class DslQueryHandler extends LoggingObject implements QueryHandler {
 
     protected String appendConstraintAndOrderByToQuery(String previousMaxConstraintColumnValue) {
         String constrainedDsl = userDslQuery;
-        if (previousMaxConstraintColumnValue != null) {
-            String constraintPhrase = ".where(op.gt(op.col('" + constraintColumnName + "'), '" + previousMaxConstraintColumnValue + "'))"
-                + ".orderBy(op.asc('" + constraintColumnName + "'))";
+        if (constraintColumnName != null) {
+            String constraintPhrase = "";
+            if (previousMaxConstraintColumnValue != null) {
+                constraintPhrase += ".where(op.gt(op.col('" + constraintColumnName + "'), '" + previousMaxConstraintColumnValue + "'))";
+            }
+            constraintPhrase += ".orderBy(op.asc('" + constraintColumnName + "'))";
             constrainedDsl = userDslQuery + constraintPhrase;
         }
         return constrainedDsl;
--- a/src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
+++ b/src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
@@ -2,6 +2,7 @@ package com.marklogic.kafka.connect.source;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.document.XMLDocumentManager;
+import com.marklogic.client.expression.PlanBuilder;
 import com.marklogic.client.io.DocumentMetadataHandle;
 import com.marklogic.client.io.FileHandle;
 import com.marklogic.client.io.StringHandle;
@@ -16,11 +17,13 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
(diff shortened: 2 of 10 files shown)


src/main/java/com/marklogic/kafka/connect/source/DslQueryHandler.java
```diff
@@ -43,10 +47,13 @@ protected String injectConstraintIntoQuery(String previousMaxConstraintColumnVal
         if (previousMaxConstraintColumnValue != null) {
             String constraintPhrase = ".where(op.gt(op.col(\"" + constraintColumnName + "\"), \"" + previousMaxConstraintColumnValue + "\"))";
             String originalDslNoWhitespace = QueryHandlerUtil.removeWhitespace(userDslQuery);
-            if (originalDslNoWhitespace.contains(").")) {
-                int firstClosingParen = originalDslNoWhitespace.indexOf(").");
-                constrainedDsl = originalDslNoWhitespace.substring(0, firstClosingParen + 1)
-                    + constraintPhrase + originalDslNoWhitespace.substring(firstClosingParen + 1);
+
+            Pattern pattern = Pattern.compile(BALANCED_PARENS_REGEX_PATTERN);
```
rjrudin, line 51: This logic is a little complex (in the sense of - I don't think it's immediately clear what's going on here), and I think a separate method would be nice, both for explaining what's going on and in case you want to unit test it directly - e.g.

    protected int findEndPositionOfFirstAccessor(originalDslNoWhitespace)

The benefit of that would be very targeted unit-testing of the very-complicated regex from stackoverflow. 

Even if it's not unit-tested directly, it's nice for self-documenting purposes and also simplifies the overall method. 

I was thinking an "Optional" would be nice here too to make it clear that something might not be found:

    protected Optional<Integer> find...

src/test/java/com/marklogic/kafka/connect/source/DslConstraintInjectionTest.java
```diff
@@ -1,61 +1,202 @@
...
+        String fromViewClause = "op.fromView('main','employees',null,op.fragmentIdCol('docId'))";
+        String userDsl = fromViewClause + ".orderBy(op.asc(\"ID\")).limit(1000)";
+        parsedConfig.put(MarkLogicSourceConfig.DSL_QUERY, userDsl);
+        String expectedResult = fromViewClause + expectedWhereClause + ".orderBy(op.asc(\"" + constraintColumn + "\")).limit(1000)";
+
+        String result = injectValueIntoQuery();
+        assertEquals(expectedResult, result, "The where clause should have been injected just after the closing paren of the fromSearch function");
+    }
+
+    private String injectValueIntoQuery() {
```
rjrudin, line 199: This can do a little more work and thus make every test simpler - you can change it to:

    private String injectValueIntoQuery(String query)

And then it will stuff the value into parsedConfig. And then every test can remove that line that stuffs it into parsedConfig. 

You could even make this method take care of everything:

    private void assertValueIsInjected(String userQuery, String expectedResult, String message)

And then have tests like this:

```
    @Test
    void fromView() {
        String fromViewClause = "op.fromView('main','employees',null,op.fragmentIdCol('docId'))";
        assertValueIsInjected(
            fromViewClause + ".orderBy(op.asc(\"ID\")).limit(1000)",
            fromViewClause + expectedWhereClause + ".orderBy(op.asc(\"" + constraintColumn + "\")).limit(1000)",
            "The where clause should have been injected just after the closing paren of the fromSearch function"
        );
    }
```

That might be a little too much, but I would at least move the parsedConfig bit into the private method.

src/main/java/com/marklogic/kafka/connect/source/QueryHandlerUtil.java
```diff
@@ -10,4 +10,6 @@ protected static String sanitize(String target) {
     protected static String removeWhitespace(String target) {
         return target.replaceAll(WHITESPACE_REGEX_PATTERN, "");
     }
+
+    protected final static String BALANCED_PARENS_REGEX_PATTERN = "(?=\\()(?:(?=.*?\\((?!.*?\\1)(.*\\)(?!.*\\2).*))(?=.*?\\)(?!.*?\\2)(.*)).)+?.*?(?=\\1)[^(]*(?=\\2$)";
```
rjrudin, line 14: I recommend including a URL for where this was pasted from in case anyone has questions about it later on. Good for citation purposes too.

src/test/java/com/marklogic/kafka/connect/source/DslConstraintInjectionTest.java
```diff
@@ -1,61 +1,202 @@
...
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 class DslConstraintInjectionTest extends AbstractIntegrationSourceTest {
-    private static final String constraintColumn = "lucky_number";
-    private static final String constraintValue = "52";
+    private static final String constraintColumn = "ID";
+    private static final String constraintValue = "2";
+    private static final String expectedWhereClause = ".where(op.gt(op.col(\"" + constraintColumn + "\"), \""+ constraintValue + "\"))";
```
rjrudin, line 18: To make it easier to read these queries, I recommend using single quotes so you don't need ugly escaped double quotes in Java. 

Also, I go back and forth on using variables for strings in tests. The reason is that it makes understanding the test a little harder - i.e. this would be easier to read, even though it duplicates some stuff:

    private static final String easierToReadClause = ".where(op.gt(op.col('ID'), '2'))";

Thoughts on that? It's a balance between duplication of values and readability of the test, there's not really a right answer.

src/main/java/com/marklogic/kafka/connect/source/DslQueryHandler.java
BillFarber: Actually, I have removed the if statement, since if the matcher doesn't find anything, that would be a runtime error.
So, I'm not using Optional here.

src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
```diff
@@ -117,4 +119,24 @@ private void assertTopicAndSingleValue(List<SourceRecord> newSourceRecords, Stri
     protected Stream<ObjectNode> recordsToJsonObjects(List<SourceRecord> records) {
         return records.stream().map(record -> readJsonObject((String) record.value()));
     }
+
+    protected String injectValueIntoQuery(String userDsl, Map<String, Object> parsedConfig, String CONSTRAINT_VALUE) {
+        parsedConfig.put(MarkLogicSourceConfig.DSL_QUERY, userDsl);
+        return new DslQueryHandler(null, parsedConfig).injectConstraintIntoQuery(CONSTRAINT_VALUE);
+    }
+
+    protected void verifyQueryReturnsExpectedRows(String constraintValue, int expectedRowCount, String stringInFirstRecord, Map<String, Object> parsedConfig) {
```
rjrudin, line 128: This seems to be recreating what the Task already does. IIUC, the intent is to verify how the task works with a simulated max constraint column value. If so, I think an approach that allows us to reuse the Task here is:

1. Add a protected setter on constraintValueStore in the Task class, and give it a comment of e.g. "Only for testing" (we have some methods like this already)
2. Create a TestConstraintValueStore that accepts the given constraintValue and set that on the Task after it's been started. 
3. Then call task.poll() like normal, and you should be able to make the same assertion on the records you get back.

src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
```diff
@@ -117,4 +119,24 @@ private void assertTopicAndSingleValue(List<SourceRecord> newSourceRecords, Stri
...
+        QueryHandler queryHandler = QueryHandler.newQueryHandler(getDatabaseClient(), parsedConfig);
+        PlanBuilder.Plan plan = queryHandler.newPlan(constraintValue);
+        PlanInvoker.Results results = PlanInvoker.newPlanInvoker(getDatabaseClient(), parsedConfig).invokePlan(plan, "Authors");
+        List<SourceRecord> newRecords = results.getSourceRecords();
+        assertEquals(expectedRowCount, newRecords.size());
+        assertTrue(((String) newRecords.get(0).value()).contains(stringInFirstRecord));
```
rjrudin, line 134: Whenever assertTrue or assertFalse is used, it's good to include a message so that when the test fails, we know what the actual value is (as JUnit won't tell us anything other than "assertion failed"). And in a helper method like this that's used across tests, it's helpful to include the value that was passed in. So do:

    String firstRecordValue = (String)newRecords.get(0).value();
    assertTrue(firstRecordValue.contains(stringInFirstRecord), 
    "Did not find " + stringInFirstRecord " + " in " + firstRecordValue);

src/test/java/com/marklogic/kafka/connect/source/DslConstraintInjectionTest.java
```diff
@@ -1,61 +1,122 @@
...
-            constraintValue + "\")).select([op.as(\"Last Name\",op.col(\"LastName\"),op.as(\'Fore Name\',op.col(\'ForeName\')])";
-        Assertions.assertEquals(expectedValue, dslQueryHandler.injectConstraintIntoQuery(constraintValue));
+        String userDsl = "op.fromView('Medical', 'Authors').select([op.as('Last Name', op.col('LastName'), op.as('Fore Name', op.col('ForeName')])";
+        String expectedValue = "op.fromView('Medical','Authors').where(op.gt(op.col('ID'), '2')).select([op.as('Last Name',op.col('LastName'),op.as('Fore Name',op.col('ForeName')])";
+        assertEquals(expectedValue, injectValueIntoQuery(userDsl, parsedConfig, "2"));
+    }
+
+    @Test
+    void wordQueryWithPhraseOrderByAndLimit() {
+        String userDsl = "op.fromDocUris(cts.wordQuery('my phrase'))";
```
rjrudin, line 49: For this test, I think it's worth ordering it by the "uri" column that comes back. I can see a user actually doing that (assuming they use URIs with a meaningful order to them), as ML does allow for ordering results based on URI order. And it's better to have a test where the constraint column exists in the rows coming back, even though we're not running the query in this test.

src/test/java/com/marklogic/kafka/connect/source/DslConstraintInjectionTest.java
```diff
@@ -1,61 +1,122 @@
...
+        String expectedResult = userDsl + ".where(op.gt(op.col('ID'), '2')).orderBy(op.asc('ID')).limit(1000)";
+        assertEquals(expectedResult, injectValueIntoQuery(userDsl + ".orderBy(op.asc('ID')).limit(1000)", parsedConfig, "2"),
+            "The where clause should have been injected just after the closing paren of the fromDocUris function");
+    }
+
+    @Test
+    void wordQueryWithPhraseAndFollowOnFunction() {
+        String userDsl = "op.fromDocUris(cts.wordQuery('my phrase').followOn('abc'))";
```
rjrudin, line 57: I know the query's not being executed, but just to make it realistic, swap out "followOn" for "joinDoc" because that's likely what a user would use next.

rjrudin (approved):

PR #122, opened 2023-01-12
DEVEXP-127: Configure source connector with key strategy

rahulvudutala:


src/main/java/com/marklogic/kafka/connect/source/CsvPlanInvoker.java
```diff
@@ -29,9 +33,12 @@ public Results invokePlan(PlanBuilder.Plan plan, String topic) {
         if (result.get() != null) {
             try (BufferedReader reader = new BufferedReader(new StringReader(result.get()))) {
                 String headers = reader.readLine();
+                AtomicLong rowNumber = new AtomicLong(0);
                 reader.lines().forEach(line -> {
+                    KeyGenerator keyGenerator = KeyGenerator.newKeyGenerator(parsedConfig, rowNumber.getAndIncrement());
+                    String key = keyGenerator != null ? keyGenerator.getUniqueKey() : null;
```
BillFarber, line 39: The KeyGenerator instance should be an instance field.
Then, the call to getUniqueKey should accept the parameters that are unique to the row, and generate the unique key based on them.
That way the KeyGenerator instance will be reusable throughout the life of the Invoker.

src/main/java/com/marklogic/kafka/connect/MarkLogicConfig.java
```diff
@@ -96,8 +96,14 @@ public boolean visible(String name, Map<String, Object> parsedConfig) {
...
-            Object uppercaseValue = value instanceof String ? ((String) value).toUpperCase() : value;
-            if (!validValues.contains(uppercaseValue)) {
+            boolean isValidValue;
+            if(value instanceof String) {
+                isValidValue = validValues.stream().map(validValue -> validValue.toString()).anyMatch(((String) value)::equalsIgnoreCase);
+            } else {
+                isValidValue = validValues.contains(value);
+            }
+
+            if (!isValidValue) {
```
BillFarber, line 106: I'm curious about this change. What was the reason for it?

src/main/java/com/marklogic/kafka/connect/source/KeyGenerator.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.Map;
+
+public interface KeyGenerator {
+    String getUniqueKey();
+
+    static KeyGenerator newKeyGenerator(Map<String, Object> parsedConfig, Long rowNumber) {
+        if(!parsedConfig.get(MarkLogicSourceConfig.KEY_STRATEGY).toString().equalsIgnoreCase("NULL")) {
+            return parsedConfig.get(MarkLogicSourceConfig.KEY_STRATEGY).toString().equalsIgnoreCase("UUID") ?
+                new UuidKeyGenerator() : new TimestampKeyGenerator(rowNumber);
```
BillFarber, line 11: Nitpick - Maybe it's just me, but with this formatting, I find it a bit tricky to notice the TimestampKeyGenerator constructor call.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -29,6 +29,9 @@ public class MarkLogicSourceConfig extends MarkLogicConfig {
     enum OUTPUT_TYPE {JSON, XML, CSV}
     private static final CustomRecommenderAndValidator OUTPUT_FORMAT_RV =
         new CustomRecommenderAndValidator(Arrays.stream(MarkLogicSourceConfig.OUTPUT_TYPE.values()).map(Enum::toString).toArray(String[]::new));
+    public static final String KEY_STRATEGY = "ml.source.key.strategy";
+    private static final CustomRecommenderAndValidator KEY_STRATEGY_RV =
+        new CustomRecommenderAndValidator("UUID", "TIMESTAMP", "NULL");
```
rjrudin, line 34: I don't think we want "NULL" as a value - we just want this to be optional, and so I believe we want the default value to be `ConfigDef.NO_DEFAULT_VALUE`.

src/main/java/com/marklogic/kafka/connect/source/TimestampKeyGenerator.java
```diff
@@ -0,0 +1,16 @@
+package com.marklogic.kafka.connect.source;
+
+import java.time.Instant;
+
+public class TimestampKeyGenerator implements KeyGenerator {
```
rjrudin, line 5: Per the story - "rowNumber is simply the index of the row in the result set" - I think the way this should work is:

1. After a PlanInvoker runs a plan, it should instantiate a KeyGenerator (KG)
2. It invokes the KG for each row
3. Thus, this class - TKG - should have a rowNumber field that defaults to 1, and it gets incremented after each call to getUniqueKey

src/main/java/com/marklogic/kafka/connect/source/KeyGenerator.java
```diff
@@ -0,0 +1,15 @@
+package com.marklogic.kafka.connect.source;
+
+import java.util.Map;
+
+public interface KeyGenerator {
+    String getUniqueKey();
```
rjrudin, line 6: "unique" is potentially misleading here - we're not offering any guarantee that the value will be unique. Just do `generateKey` so that we don't imply something that may not be true (for example, if we support column-based keys, it's on the user to determine if the keys will be unique or not - and they may intentionally not want them to be unique).

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
BillFarber, line 34: I think that if "NULL" is going to be an actual strategy, then we should have an implementation for it.
Also, I wonder if we can come up with a better name, so that "NULL" isn't overloaded.

BillFarber (changes requested):
I haven't finished looking at the test changes yet. I'll get to that in an hour or so.

src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
```diff
@@ -89,8 +92,13 @@ void loadSingleAuthorRowIntoMarkLogicWithCustomTime(String uri, String id, Strin
     }
 
     void verifyQueryReturnsFifteenAuthors(List<SourceRecord> sourceRecords, String expectedValue) {
+        verifyQueryReturnsFifteenAuthors(sourceRecords, expectedValue, "null");
```
rjrudin, line 95: Just do `null` here instead of the string - `"null"` implies to me that it's something besides `null`, but it's just `null`.

src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
```diff
@@ -108,6 +116,21 @@ private void assertTopicAndSingleValue(List<SourceRecord> newSourceRecords, Stri
             "List of SourceRecords does not contain a record with the expected value");
     }
 
+    private void assertRecordKeys(List<SourceRecord> newSourceRecords, String keyStrategy) {
+        if(keyStrategy.equalsIgnoreCase("uuid")) {
+            newSourceRecords.forEach(sourceRecord -> Assertions.assertEquals(UUID.fromString(sourceRecord.key().toString()).toString(), sourceRecord.key().toString()));
```
rjrudin, line 121: Use `assertEquals` instead of `Assertions.assertEquals` - see https://wiki.marklogic.com/display/PM/Writing+Tests+for+Developer+Experience for a tip on how to do this easily in Intellij.

rjrudin (changes requested):

src/main/java/com/marklogic/kafka/connect/source/CsvPlanInvoker.java
rjrudin: I think we only want it to be reused throughout a single `invokePlan`. While the UUID one could be reused throughout the life of the invoker, the Timestamp one cannot be since it includes the row number. And for each `invokePlan` call, we'd want the first row to have a row number of 1.

src/main/java/com/marklogic/kafka/connect/source/KeyGenerator.java
rjrudin: What I do when using "?" is: 

```
return something that returns true or false ? 
    this thing :
    this other thing;
```

I find that usually reads pretty well.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
rjrudin: I like this - a `NullKeyGenerator` makes sense and means that whatever calls the `KeyGenerator` doesn't need to know about the concept of `null` as a key.

I'm not sure though that "NULL" should be a choice for the user. I feel like making the field optional works, and the absence of a user choice means we use `NullKeyGenerator`. Though - I can see the benefit of making this explicit, but I wonder if "None" would be a little less developer-speakish.

BillFarber: Yep, I think "None" is good.

src/main/java/com/marklogic/kafka/connect/source/CsvPlanInvoker.java
```diff
@@ -29,9 +33,11 @@ public Results invokePlan(PlanBuilder.Plan plan, String topic) {
         if (result.get() != null) {
             try (BufferedReader reader = new BufferedReader(new StringReader(result.get()))) {
                 String headers = reader.readLine();
+                AtomicLong rowNumber = new AtomicLong(1);
```
rjrudin, line 36: Java is annoying here, and the use of AtomicLong is more complex than it should be - I think it's better to convert the "forEach" call to just be a regular for loop, and then use a regular "int rowNumber". The performance difference between "rowNumber++" and "rowNumber.getAndIncrement()" is likely very close to zero - it's more the simplicity that I'm looking for.

src/main/java/com/marklogic/kafka/connect/source/KeyGenerator.java
```diff
@@ -0,0 +1,19 @@
+package com.marklogic.kafka.connect.source;
+
+import java.util.Map;
+
+public interface KeyGenerator {
+    String generateKey(Long rowNumber);
```
rjrudin, line 6: Just use "long" here - "Long" implies that the value could be null.

src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConfig.java
```diff
@@ -63,7 +66,10 @@ private static ConfigDef getConfigDef() {
             .define(WAIT_TIME, Type.LONG, 5000, ConfigDef.Range.atLeast(0), Importance.MEDIUM,
                 "Required, must be zero or higher; the amount of time in milliseconds to wait, on each poll call, " +
                     "before querying MarkLogic. Kafka will continually call poll on the source connector, so this " +
-                    "can be used to control how frequently the connector queries MarkLogic.");
+                    "can be used to control how frequently the connector queries MarkLogic.")
+            .define(KEY_STRATEGY, Type.STRING, "NONE", KEY_STRATEGY_RV, Importance.MEDIUM,
+                "Assigns a key for each record returned; either 'UUID', 'TIMESTAMP', or 'NONE'. If not set key is not returned for the records",
```
rjrudin, line 71: For the last sentence, say - "If not set, then NONE will be used, in which case a key is not created for each record."

src/main/java/com/marklogic/kafka/connect/source/TimestampKeyGenerator.java
```diff
@@ -0,0 +1,11 @@
...
+
+    @Override
+    public String generateKey(Long rowNumber) {
+        return Long.toString(Instant.now().toEpochMilli()).concat("-").concat(Long.toString(rowNumber));
```
rjrudin, line 9: Is this the same as `System.currentTimeInMillis()`? 

@BillFarber  It occurs to me now that perhaps the ML server timestamp would be more helpful here. The reason being that I think it would be preferable if each of these keys have the exact same timestamp, with only the row number differing. And the key now has some meaning - it's tied to something in MarkLogic (though that something - the server timestamp - unfortunately is not durable and will disappear at some point in the future).

src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
```diff
@@ -92,25 +94,45 @@ void loadSingleAuthorRowIntoMarkLogicWithCustomTime(String uri, String id, Strin
...
-        Assertions.assertTrue(foundExpectedValue.get(),
+        assertTrue(foundExpectedValue.get(),
             "List of SourceRecords does not contain a record with the expected value");
     }
 
+    private void assertRecordKeys(List<SourceRecord> newSourceRecords, String keyStrategy) {
+        if(keyStrategy.equalsIgnoreCase("uuid")) {
+            newSourceRecords.forEach(sourceRecord -> assertEquals(UUID.fromString(sourceRecord.key().toString()).toString(), sourceRecord.key().toString()));
+        } else if(keyStrategy.equalsIgnoreCase("timestamp")) {
+            AtomicLong rowNumber = new AtomicLong(1);
```
rjrudin, line 125: This becomes a more interesting assertion if we use the ML server timestamp, as we could verify that each key begins with the same timestamp.

src/main/java/com/marklogic/kafka/connect/MarkLogicConfig.java
rjrudin: @rahulvudutala  I commented this change out locally and all the tests still passed. So it doesn't seem like this change is solving a problem.

src/main/java/com/marklogic/kafka/connect/source/TimestampKeyGenerator.java
BillFarber: I'm fine with using the server timestamp.  It would help identify batches ... if that would be of any use.

rjrudin: @rahulvudutala I think this works for now, I'll take a look later on to see if we can use the server timestamp instead.

src/main/java/com/marklogic/kafka/connect/source/CsvPlanInvoker.java
rjrudin: On second thought - this is fine, can leave as is.

src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
rjrudin: No action needed for now.

src/main/java/com/marklogic/kafka/connect/MarkLogicConfig.java
rahulvudutala: Reverted the code.

rjrudin (approved):

PR #124, opened 2023-01-13
prototypeLimitConfig

BillFarber:
I haven't really thought about if we need new tests for these changes.


src/main/java/com/marklogic/kafka/connect/source/SerializedQueryHandler.java
```diff
@@ -34,26 +35,50 @@ public SerializedQueryHandler(DatabaseClient databaseClient, Map<String, Object>
...
-    protected String injectConstraintIntoQuery(String previousMaxConstraintColumnValue) {
-        String constrainedSerialized = userSerializedQuery;
-        if (previousMaxConstraintColumnValue != null) {
-            String constraintNodeString = "{\"ns\":\"op\", \"fn\":\"where\", \"args\":[{\"ns\":\"op\", \"fn\":\"gt\", \"args\":[{\"ns\":\"op\", \"fn\":\"col\", \"args\":[\"" + constraintColumnName + "\"]}, \"" + previousMaxConstraintColumnValue + "\"]}]}";
-            try {
-                ObjectNode constraintNode = (ObjectNode) mapper.readTree(constraintNodeString);
+    protected String appendLimitToQuery(String currentSerializedQuery) {
+        String limitNodeString = "{\"ns\": \"op\", \"fn\": \"limit\", \"args\": [" + rowLimit + "]}";
+        try {
+            ObjectNode limitNode = (ObjectNode) mapper.readTree(limitNodeString);
```
rjrudin, line 54: Not worth changing now, but I think it would read a bit nicer to just build this JSON manually since it's simple enough, and it avoids any error from having to call readTree. Again, out of scope for this PR, just something to consider. The problem is only that Java is ugly with JSON strings - this would be much easier to read if we were writing JavaScript.

rjrudin (approved):
LGTM - I noted that the JSON strings in Java are ugly, but that can be addressed later, if ever.

BillFarber:
What do you think of this? I think it's probably better. At least now there's less duplication and it might even be more readable.

src/main/java/com/marklogic/kafka/connect/source/DslQueryHandler.java
```diff
@@ -29,27 +30,29 @@ public DslQueryHandler(DatabaseClient databaseClient, Map<String, Object> parsed
...
-            String constraintPhrase = ".where(op.gt(op.col(\"" + constraintColumnName + "\"), \"" + previousMaxConstraintColumnValue + "\"))";
-            String originalDslNoWhitespace = QueryHandlerUtil.removeWhitespace(userDslQuery);
-            if (originalDslNoWhitespace.contains(").")) {
-                int firstClosingParen = originalDslNoWhitespace.indexOf(").");
-                constrainedDsl = originalDslNoWhitespace.substring(0, firstClosingParen + 1)
-                    + constraintPhrase + originalDslNoWhitespace.substring(firstClosingParen + 1);
-            } else {
-                constrainedDsl = userDslQuery + constraintPhrase;
-            }
+            String constraintPhrase = ".where(op.gt(op.col('" + constraintColumnName + "'), '" + previousMaxConstraintColumnValue + "'))"
```
rjrudin, line 53: Just occurred to me that we can use bindParam, as that's exposed on the RawPlan interface. That allows us to avoid string concatenation of values (but as mentioned in chat, doesn't seem possible yet to avoid string concatenation entirely). I think this should do the trick:

```
PlanBuilder.Plan plan = databaseClient.newRowManager().newRawQueryDSLPlan(new StringHandle(currentDslQuery));
        if (rowLimit > 0) {
            plan.bindParam("rowLimit", rowLimit);
        }
        if (previousMaxConstraintColumnValue != null) {
            plan.bindParam("previousMaxConstraintColumnValue", previousMaxConstraintColumnValue);
        }
        return plan;
```

That should work after modifying the query to do something like this:

    return currentDslQuery + ".limit(op.param('rowLimit'))";

src/main/java/com/marklogic/kafka/connect/source/SerializedQueryHandler.java
```diff
@@ -96,6 +115,65 @@ private String buildMaxValueSerializedQuery() throws JsonProcessingException {
...
+                .add(mapper.createObjectNode()
+                    .put("ns", "op")
+                    .put("fn", "gt")
+                    .set("args", mapper.createArrayNode()
+                        .add(mapper.createObjectNode()
+                            .put("ns", "op")
+                            .put("fn", "col")
+                            .set("args", mapper.createArrayNode().add(constraintColumnName))
+                        )
+                        .add(previousMaxConstraintColumnValue)
```
rjrudin, line 172: I realize now that this has the added benefit of avoiding any issues with concatenating in a value.

PR #125, opened 2023-01-16
Sanitizing the previous max value

rjrudin:
Couple other changes:

1. Combined the methods in DslQueryHandler for modifying the user query into `constrainUserQuery` so that the unit test only needs to invoke one method instead of two when rowLimit is specified
2. Changed info-level logging that I'd been using for debugging to debug-level logging, particularly since it involves user data that they may not want logged
3. For consistency in the DSL query, added op.col in a couple places so that we always use op.col to refer to a column (I'm a little paranoid about this from my time with Optic Update)


rjrudin:
Closing temporarily until other PR is merged

PR #128, opened 2023-01-16
Added docs for rowLimit

rjrudin:
I renamed "row.limit" to "rowLimit" as I think the latter is consistent with the existing "outputFormat" option. Also collapsed the `appendLimitToQuery` method as it was only used in one place (I think this was formerly used by a test, but no longer is).

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ The following optional properties can also be configured:
 
 - `ml.source.waitTime` = amount of time, in milliseconds, that the connector will wait on each poll before running the Optic query; defaults to 5000
 - `ml.source.optic.outputFormat` = the format of rows returned by the Optic query; defaults to "JSON", and can instead be "XML" or "CSV"
+- `ml.source.optic.rowLimit` = the maximum number of rows to retrieve per poll
 - `ml.source.optic.constraintColumn.name` = name of a column returned by the Optic query to use for constraining results on subsequent runs
 - `ml.source.optic.constraintColumn.uri` = URI of a document that the connector will write after each run that uses a constraint column; the document will contain the highest value in the constraint column
 - `ml.source.optic.constraintColumn.collections` = comma-separated list of collection names to assign to the document identified by the URI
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -299,27 +296,30 @@ column is configured, the connector will perform the on each run:
...
+2. The column should ideally contain unique values. For example, if the column contains numbers, an initial 
    query may find that the highest value in the column is 50. If it is possible for future rows to be inserted with a 
    value of 50, those rows will not be returned because the connector will constrain on rows with a value greater than 50.
    Note that a "last updated" dateTime column does not need to contain unique values, assuming that any new row would 
    always have a value greater than any existing row. 
-2. A query can make use of the Optic `limit()` modifier to limit how many rows are returned in conjunction with a 
-   constraint column as well. This may be useful in situations where the number of "new" rows can be high enough to cause
-   the Kafka Connect process to run out of memory. When doing so though, the user will typically want to order the 
-   results by the same column used as the constraint column to ensure that no rows are missed. 
+3. If using the row limit feature described above, the rows should be sorted in ascending order based on the
```
BillFarber, line 309: > the rows should be sorted in ascending order
This will happen automatically now.

README.md
```diff
@@ -299,27 +296,30 @@ column is configured, the connector will perform the on each run:
...
-   the Kafka Connect process to run out of memory. When doing so though, the user will typically want to order the 
-   results by the same column used as the constraint column to ensure that no rows are missed. 
+3. If using the row limit feature described above, the rows should be sorted in ascending order based on the 
+   constraint column. This ensures that no rows will be skipped. 
    
-Consider the following example of using a constraint column and a limit together:
+Consider the following example of using a constraint column with a row limit:
 
-    ml.source.optic.query=op.fromView('demo', 'purchases').where(op.sqlCondition('price_per_unit > 10')).orderBy(op.asc('id')).limit(1000)
+    ml.source.optic.query=op.fromView('demo', 'purchases').where(op.sqlCondition('price_per_unit > 10')).orderBy(op.asc(op.col('id'))
```
BillFarber, line 314: The user query should not include the .where or .orderBy clauses.

BillFarber (changes requested):

README.md
rjrudin: This "where" clause is unrelated and can be kept - I'll remove the "orderBy" clause and make it clear that it'll be included.

BillFarber (approved):

PR #130, opened 2023-01-17
Throwing error when max value query fails

rjrudin:
Both the DSL and serialized handlers do the same thing now. Tried a base class for providing the common behavior, but that felt a little over-engineered, so resurrected QueryHandlerUtil and added a static convenience method there instead. 

Also modified the catch block in MarkLogicConstraintValueStore to catch any exception and throw the same error. 

For testing, I had to manually force an error in order to get executeMaxValuePlan to fail - verified that the error message is correctly constructed.


BillFarber (approved):
LGTM

PR #136, opened 2023-01-18
Moved Kafka-specific classes out of client.ext package

rjrudin:
The IdStrategy is Kakfa-specific, and DocumentWriteOperationBuilder can be used for inspiration for something reusable in the Java Client.


BillFarber (approved):
Nice

PR #137, opened 2023-01-18
Using JsonNode in serialized handler

rjrudin:
Intent is to reduce the number of areas where Jackson can throw an exception. There's now just one place - when the handler is created, the String query is converted into a JsonNode, and then everything after that deals with the JsonNode.


BillFarber (approved):
Nice. I should have thought of doing this earlier.

PR #152, opened 2023-02-03
Added support for MarkLogic Cloud auth

rjrudin:
Fairly simple matter of just adding two config options and passing them along to the existing class that collects inputs for constructing a client.


BillFarber (approved):
LGTM

PR #153, opened 2023-02-06
Little bits of beta feedback

rjrudin:
Summary:

1. Bumped to ml-javaclient/app-deployer 4.5.0, which uses Java Client 6.1.0
2. Clarified why a user may have a serialized plan in the README (I really think the Java Client is the main use case here; if a user is using qconsole, wouldn't they just use a DSL?).
3. Modified the source connector to ignore a task count greater than 1.

### DIFF
--- a/src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConnector.java
+++ b/src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConnector.java
@@ -71,10 +71,13 @@ public class MarkLogicSourceConnector extends SourceConnector {
 
     @Override
     public List<Map<String, String>> taskConfigs(final int taskCount) {
-        final List<Map<String, String>> configs = new ArrayList<>(taskCount);
-        for (int i = 0; i < taskCount; ++i) {
-            configs.add(config);
+        if (taskCount > 1) {
+            logger.warn("As of the 1.8.0 release, the Kafka tasks.max property is ignored and a single source connector " +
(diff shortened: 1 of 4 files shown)


src/main/java/com/marklogic/kafka/connect/source/MarkLogicSourceConnector.java
```diff
@@ -71,10 +71,13 @@ public Class<? extends Task> taskClass() {
 
     @Override
     public List<Map<String, String>> taskConfigs(final int taskCount) {
-        final List<Map<String, String>> configs = new ArrayList<>(taskCount);
-        for (int i = 0; i < taskCount; ++i) {
-            configs.add(config);
+        if (taskCount > 1) {
```
rjrudin, line 74: @BillFarber  @bgeorge111  This is what the user sees now if they set tasks.max to 2 or higher:

[2023-02-06 15:06:38,023] INFO [marklogic-employees-source|task-0] Starting RowManagerSourceTask (com.marklogic.kafka.connect.source.RowManagerSourceTask:67)
[2023-02-06 15:06:38,024] WARN [marklogic-employees-source|worker] As of the 1.8.0 release, the Kafka tasks.max property is ignored and a single source connector task is created. This prevents duplicate records from being created via multiple instances of the task with the exact same config. (com.marklogic.kafka.connect.source.MarkLogicSourceConnector:75)

rjrudin: And the README has been updated to mention this as well.

PR #154, opened 2023-02-10
DHF integration in sink connector now works with Confluent

rjrudin:
See the comment in build.gradle about excluding Jackson dependencies from the connector archive. 

Bumped up a couple test dependencies while I was at it.


BillFarber (approved):

rjrudin:
There's one test failure that started happening yesterday for an unrelated change. Will investigate on the develop branch.

PR #155, opened 2023-02-10
Improving error message for when max value query fails

rjrudin:
We were catching the error when the query succeeds but brings back an unexpected result. We now catch the error when the query fails. This is in response to the `pullDataUsingFromSearchDocs` test failing due to a change in the server where the "select" call fails when used with "fromSearchDocs".

### DIFF
--- a/src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
+++ b/src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
@@ -41,6 +41,7 @@ import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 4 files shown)


src/test/java/com/marklogic/kafka/connect/source/AbstractIntegrationSourceTest.java
```diff
@@ -156,6 +157,7 @@ protected void verifyQueryReturnsExpectedRows(String constraintValue, int expect
         task.setConstraintValueStore(new TestConstraintValueStore((String)parsedConfig.get(MarkLogicSourceConfig.CONSTRAINT_COLUMN_NAME), constraintValue));
         try {
             List<SourceRecord> newRecords = task.poll();
+            assertNotNull(newRecords, "poll() unexpectedly returned null; params: " + params + "; constraintValue: " + constraintValue);
```
rjrudin, line 160: Added this so that the test fails with a nicer error message instead of an NPE.

BillFarber (approved):

PR #160, opened 2023-03-14
Added test for writing Avro data

rjrudin:
Not updating docs yet, as I'm not clear exactly on all the ways that Avro data can arrive. This at least gives us confidence that a byte array generated by Avro can be written as JSON.


BillFarber (approved):
LGTM

PR #163, opened 2023-06-10
Added Gradle tasks for installing the ML Connector and the Datagen connector in a local CP install

BillFarber:


build.gradle
```diff
@@ -235,6 +235,22 @@ task copyConnectorToConfluent(type: Copy, group: confluentTestingGroup, dependsO
...
+//  1) It uses $CONFLUENT_HOME to determine which CP installation to use
+//  2) It automatically uninstalls any current version of the connector
+//  3) It automatically adds the "confluent-hub-components" directory to all known worker properties files
+task installConnectorInConfluent(type: Exec, group: confluentTestingGroup, dependsOn: [connectorArchive]) {
+  description = "Uses 'Confluent-hub' to install the connector in your local Confluent Platform"
+  commandLine "confluent-hub", "install", "--no-prompt", "build/distro/${baseArchiveName}.zip"
+  ignoreExitValue = true
+}
+
+task installDatagenConnectorInConfluent(type: Exec, group: confluentTestingGroup, dependsOn: [connectorArchive]) {
```
rjrudin, line 248: This is a one-time task that's already documented in the CONTRIB file - what value does a Gradle task add for it?

rjrudin (changes requested):
This seems to be the better way to install locally, but if so, the CONTRIB needs to be updated and it seems likely we also need `copyConnectorToConfluent` removed, and possibly `confluentHome` removed as well. And hopefully `deleteConnectorInConfluent` can be removed too?

rjrudin (changes requested):
This is working well for me - I verified I could change the connector and run `clean` plus the new task and it all worked - just need to find/replace the 4 occurrences of `copyConnectorToConfluent` with `installConnectorInConfluent`, and then merge away.

PR #174, opened 2024-05-08
Updating all dependencies

BillFarber:
Still need to update DH - that will be a later PR.
I'm going to move Confluent Platform to a Docker cluster first.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ plugins {
   id "com.github.jk1.dependency-license-report" version "1.19"
 
   // Only used for testing
-  id 'com.marklogic.ml-gradle' version '4.6.0'
+  id 'com.marklogic.ml-gradle' version '4.7.0'
   id 'jacoco'
   id "org.sonarqube" version "4.4.1.3373"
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -31,20 +31,22 @@ configurations {
...
-  compileOnly "org.slf4j:slf4j-api:1.7.36"
+  compileOnly "org.slf4j:slf4j-api:2.0.13"
 
   implementation 'com.marklogic:ml-javaclient-util:4.7.0'
   // Force DHF to use the latest version of ml-app-deployer, which minimizes security vulnerabilities
-  implementation "com.marklogic:ml-app-deployer:4.6.0"
+  implementation "com.marklogic:ml-app-deployer:4.7.0"
 
-  implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.15.2"
+  implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.17.1"
```
rjrudin, line 49: The Java Client is only on 2.15.3 - stick with that, particularly as I believe Kafka provides an isolated class loader for a plugin so we won't be impacted by whatever Jackson version Kafka is using (not certain about that, but think it's true).

build.gradle
```diff
@@ -63,13 +65,13 @@ dependencies {
...
-  testImplementation 'net.mguenther.kafka:kafka-junit:3.5.1'
+  testImplementation 'net.mguenther.kafka:kafka-junit:3.6.0'
 
-  testImplementation "org.apache.avro:avro-compiler:1.11.1"
+  testImplementation "org.apache.avro:avro-compiler:1.11.3"
 
   // Forcing logback to be used for test logging
-  testImplementation "ch.qos.logback:logback-classic:1.3.5"
-  testImplementation "org.slf4j:jcl-over-slf4j:1.7.36"
+  testImplementation "ch.qos.logback:logback-classic:1.5.6"
```
rjrudin, line 73: Logback 1.4 requires Java 11 or higher, so stick with 1.3.x. Verified by running the tests locally on Java 8, and logback causes an error.

build.gradle
```diff
@@ -63,13 +65,13 @@ dependencies {
...
+  testImplementation "org.slf4j:jcl-over-slf4j:2.0.13"
```
rjrudin, line 74: TIL that slf4j actually only requires Java 8, so this is a good upgrade to do in all of our projects.

rjrudin (changes requested):

build.gradle
BillFarber: sorry.

PR #176, opened 2024-05-10
Merging into a single test-app

BillFarber:


test-app/build.gradle
```diff
@@ -0,0 +1,59 @@
+plugins {
+  id 'net.saliman.properties' version '1.5.2'
+  id "com.marklogic.ml-data-hub" version "6.0.0"
+}
+
+ext {
```
rjrudin, line 6: I think you can drop this group - now that you have a simple Gradle file for the test-app, the group isn't buying us anything.

test-app/dockerMinimal/docker-compose.yml
rjrudin, line 1: For naming consistency, do "docker-minimal".

test-app/gradle.properties
```diff
@@ -0,0 +1,16 @@
+mlHost=localhost
+# Define these in gradle-local.properties
+mlUsername=
+mlPassword=
+
+mlStagingForestsPerHost=1
+mlFinalForestsPerHost=1
```
rjrudin, line 7: Bump this and staging to 2 - just to make sure we don't introduce any bugs that don't appear when there's only one forest. Jobs can remain at 1 forest, as we don't care about jobs data at all.

test-app/gradle.properties
```diff
@@ -0,0 +1,16 @@
+mlHost=localhost
+# Define these in gradle-local.properties
+mlUsername=
```
rjrudin, line 3: These can both be admin/admin now right, since we're assuming the use of Docker?

test-app/gradle.properties
```diff
@@ -0,0 +1,16 @@
...
+mlPassword=
+
+mlStagingForestsPerHost=1
+mlFinalForestsPerHost=1
+mlJobForestsPerHost=1
+
+# For the Confluent Connector Archive
+componentOwner=marklogic
```
rjrudin, line 11: These seem like they should be in the top-level gradle.properties, not part of the test-app.

rjrudin (changes requested):
Looks good, just a few small things.

PR #185, opened 2024-06-06
Update the new snapshot version

BillFarber:
Also added a little publishing information to the CONTRIBUTING file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -318,3 +318,7 @@ project. You must have Ruby installed. Additionally, there seems to be a bug wit
 The server needs to be run with Ruby 3.2.3, so you will need to run `chruby ruby-3.2.3` before starting the jekyll
 server. To start the jekyll server, cd into the /docs directory and run the command `bundle exec jekyll server`.
 This will start the server and the user documents will be available at http://127.0.0.1:4000/.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -318,3 +318,8 @@ project. You must have Ruby installed. Additionally, there seems to be a bug wit
 The server needs to be run with Ruby 3.2.3, so you will need to run `chruby ruby-3.2.3` before starting the jekyll
 server. To start the jekyll server, cd into the /docs directory and run the command `bundle exec jekyll server`.
 This will start the server and the user documents will be available at http://127.0.0.1:4000/.
+
+# Publishing the Connector to Confluent
+
+Please https://wiki.marklogic.com/display/PM/Updating+the+Confluent+release+of+the+Kafka+connector for information
```
rjrudin, line 324: This should be a Progress Wiki URL - though I think it's better to omit the URL and just refer the reader to the internal Wiki page for the connector.

rjrudin (changes requested):

CONTRIBUTING.md
BillFarber: Updated.

rjrudin (approved):

PR #191, opened 2024-10-31
Gracefully handle problems with older versions of Kafka.

BillFarber:


src/main/java/com/marklogic/kafka/connect/sink/WriteBatcherSinkTask.java
```diff
@@ -101,20 +102,34 @@ protected void writeSinkRecord(SinkRecord sinkRecord) {
...
-            originalRecord.headers().add(MARKLOGIC_MESSAGE_FAILURE_HEADER, getBytesHandleNull(failureHeaderValue));
-            originalRecord.headers().add(MARKLOGIC_MESSAGE_EXCEPTION_MESSAGE, getBytesHandleNull(e.getMessage()));
-            originalRecord.headers().add(MARKLOGIC_ORIGINAL_TOPIC, getBytesHandleNull(sinkRecord.topic()));
-            if (writeEvent != null) {
-                originalRecord.headers().add(MARKLOGIC_TARGET_URI, writeEvent.getTargetUri().getBytes(StandardCharsets.UTF_8));
+            try {
+                ConsumerRecord<byte[], byte[]> originalRecord = ((InternalSinkRecord) sinkRecord).context().original();
+                addFailureHeadersToOriginalSinkRecord(originalRecord, e, failureHeaderValue, writeEvent);
+            } catch (NoSuchMethodError methodException) {
+                throw new MarkLogicConnectorException(
```
rjrudin, line 109: I was thinking we just want a warning here - and perhaps call `addFailureHeadersToNonInternalSinkRecord` instead? Would that make sense? 

I don't think an error is good here because it will obscure the actual problem, which is whatever error caused this message to arrive here. So I think logging a warning and then optionally calling the other method is the way to go.

rjrudin (changes requested):

anu3990 (approved):

rjrudin (approved):

PR #203, opened 2025-08-04
Force v3.18 of commons-lang to fix CVE

BillFarber:


stevebio (approved):
Looks ok.

rjrudin (approved):
Note that there are a few dozen more issues reported by Black Duck. Can you open a ticket to address those as well? We'll need all of them addressed or to have an exception for them in order to get a release out with this fix.

rjrudin (approved):

PR #207, opened 2025-08-06
MLE-23377 - Bump Kafka to 3.9.1

BillFarber:
Also, forced the latest commons-collection4 and upgraded to the latest logback version.

This leaves a single "medium" security risk (spring-framework) that will be addressed next.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,17 +27,23 @@ configurations {
   documentation
   assets
 
-  // Force v3.18 of commons-lang3 to avoid security vulnerabilities, without also
-  // upgrading ml-app-deployer to 6.0.0, which we are not ready to do yet.
   configurations.all {
     resolutionStrategy {
-        force 'org.apache.commons:commons-lang3:3.18.0'
+      // Force v4.5.0 of commons-collections4 to avoid CVEs in v4.4.0 from transitive dependecies:
+      // CVE-2025-48924 (https://www.cve.org/CVERecord?id=CVE-2025-48924) and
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -71,7 +77,7 @@ dependencies {
   testImplementation "org.apache.avro:avro-compiler:1.12.0"
 
   // Forcing logback to be used for test logging
-  testImplementation "ch.qos.logback:logback-classic:1.3.14"
+  testImplementation "ch.qos.logback:logback-classic:1.5.18"
```
rjrudin, line 80: This requires Java 17 - which is fine if you're going in that direction anyway - i.e. forcing marklogic-junit5 to use Spring 6.x.

build.gradle
```diff
@@ -37,7 +37,13 @@ configurations {
 }
 
 ext {
-  kafkaVersion = "3.8.1"
+  kafkaVersion = "3.9.1"
+
+  configurations.all {
+    resolutionStrategy {
+      force "org.apache.commons:commons-collections4:4.5.0"
```
rjrudin, line 44: Add a comment here to explain why this is being done - i.e. what CVE is being addressed due to which earlier version of this library.

rjrudin (changes requested):

build.gradle
BillFarber: I think that forcing Java 17 for testing is fine. So, yes, that's the plan.

rjrudin (approved):

PR #208, opened 2025-08-07
MLE-23405 - Force Spring Framework to v6 in tests.

BillFarber:
This also forces the use of Java 17 for tests.

This clears up all the Spring Framework risks in BlackDuck.


BillFarber:
Closing since Spring is also a dependency of ml-app-deployer, which would force the runtime to also be Java 17 and we don't want that yet.

PR #222, opened 2025-10-01
MLE-24529 - Upgrade to the latest MarkLogic 12 Docker image.

BillFarber:


.env
```diff
@@ -1 +1,2 @@
-MARKLOGIC_IMAGE=ml-docker-db-dev-tierpoint.bed-artifactory.bedford.progress.com/marklogic/marklogic-server-ubi:latest-12
+MARKLOGIC_IMAGE=ml-docker-db-dev-tierpoint.bed-artifactory.bedford.progress.com/marklogic/marklogic-server-ubi-rootless:12.0.0-ubi-rootless-2.2.2
```
rjrudin, line 1: Breaking news - this actually the tag to use 

ml-docker-db-dev-tierpoint.bed-artifactory.bedford.progress.com/marklogic/marklogic-server-ubi-rootless:12.1.nightly-ubi-rootless

Jenkinsfile
```diff
@@ -6,37 +6,52 @@ pipeline{
     buildDiscarder logRotator(artifactDaysToKeepStr: '7', artifactNumToKeepStr: '', daysToKeepStr: '30', numToKeepStr: '')
   }
   environment{
-    JAVA_HOME_DIR="/home/builder/java/jdk-17.0.2"
+    JAVA17_HOME_DIR="/home/builder/java/jdk-17.0.2"
     GRADLE_DIR   =".gradle"
     DMC_USER     = credentials('MLBUILD_USER')
     DMC_PASSWORD = credentials('MLBUILD_PASSWORD')
+    MARKLOGIC12_IMAGE_TAG = "ml-docker-db-dev-tierpoint.bed-artifactory.bedford.progress.com/marklogic/marklogic-server-ubi-rootless:12.0.0-ubi-rootless-2.2.2"
```
rjrudin, line 13: I don't think we want this yet because it's suggesting that it's a tag other than what's in .env. Until we have a use case for running the tests against multiple MarkLogic version, leave this out so that the value in .env is used.

build.gradle
```diff
@@ -115,23 +115,28 @@ tasks.withType(JavaCompile) {
...
-    // xml.enabled true
+  jacocoTestReport {
+    dependsOn test
+    reports {
+      xml.required = true
+    }
   }
-}
 
+  // Turning off coverage verification, as we haven't figured out how to configure Sonar correctly to pick up the jacoco
```
rjrudin, line 126: This comment isn't true - it was relevant for the Spark connector which had a separate code-coverage-report project, but that's not the case here. Should be able to either remove this or set it to a value slightly lower than whatever the current code coverage is.

rjrudin (changes requested):

PR #225, opened 2025-10-03
MLE-24529 - Upgrades Java and Spring versions (based on Copilot's suggestion)

BillFarber:
NOTE - This PR is an either/or with PR #226. I'll wait to see comments to decide which one to use.

Upgrades Java to version 17 and Spring to version 6.

This change upgrades the Java source and target compatibility to version 17, and enforces Spring 6 dependencies to address potential CVEs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,8 +27,8 @@ plugins {
 }
 
 java {
-  sourceCompatibility = 1.8
-  targetCompatibility = 1.8
+  sourceCompatibility = JavaVersion.VERSION_17
+  targetCompatibility = JavaVersion.VERSION_17
 }
 
 repositories {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -44,12 +44,19 @@ configurations {
...
-      // upgrading ml-app-deployer to 6.0.0, which we are not ready to do yet.
-      force 'org.apache.commons:commons-lang3:3.18.0'
+      eachDependency { DependencyResolveDetails details ->
+        if (details.requested.group == 'org.apache.commons' && details.requested.name == 'commons-collections4') {
+          details.useVersion '4.5.0'
+        }
+        if (details.requested.group == 'org.apache.commons' && details.requested.name == 'commons-lang3') {
+          details.useVersion '3.18.0'
+        }
+        if (details.requested.group == 'org.springframework') {
```
rjrudin, line 54: The problem with this is the reader has no idea what's bringing in Spring. We only have one problem - which is that marklogic-junit5 is bring in Spring 5. So it's better to make that problem explicit by modifying the marklogic-junit5 dependency until it's been bumped up to use Spring 6.

build.gradle
```diff
@@ -44,12 +44,19 @@ configurations {
       // Force v4.5.0 of commons-collections4 to avoid CVEs in v4.4.0 from transitive dependecies:
       // CVE-2025-48924 (https://www.cve.org/CVERecord?id=CVE-2025-48924) and
       // CVE-2020-15250 (https://www.cve.org/CVERecord?id=CVE-2020-15250)
-      force "org.apache.commons:commons-collections4:4.5.0"
-
-      // Force v3.18 of commons-lang3 to avoid CVE-2025-48924
-      // (https://www.cve.org/CVERecord?id=CVE-2025-48924), without also
-      // upgrading ml-app-deployer to 6.0.0, which we are not ready to do yet.
-      force 'org.apache.commons:commons-lang3:3.18.0'
+      eachDependency { DependencyResolveDetails details ->
```
rjrudin, line 47: You can hopefully get rid of these once you've upgraded to Kafka 4. Per the comment on the commons-lang3 one - that was due to an older version of ml-app-deployer, so it's likely that that "force" can be removed. Note as well, we never want to have these "useVersion" blocks in here without a corresponding `details.because` to document why we're forcing a version.

rjrudin (commented):
Yeah don't do this one.

PR #228, opened 2025-10-07
MLE-24529 - Force newer jetty and netty libraries to avoid CVEs

BillFarber:
org.eclipse.jetty:jetty-http:12.1.1
io.netty:netty-all:4.2.6.Final

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,6 +54,25 @@ configurations {
       // Force v3.19 of commons-lang3 to avoid CVE-2025-48924 (https://www.cve.org/CVERecord?id=CVE-2025-48924), which
       // is caused by the use of avro-compiler v1.12.0 with older dependencies including commons-lang3 v3.12.0.
       force 'org.apache.commons:commons-lang3:3.19.0'
+
+      resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+        // Force v12.1.1 of jetty-http to avoid CVE-2025-5115
+        // (https://nvd.nist.gov/vuln/detail/CVE-2025-5115), which is a transitive
+        // dependency of Kafka connect-runtime:4.1.0
+        // Need to ensure this inclusdes all jetty modules, such as "org.eclipse.jetty.ee10"
+        if (details.requested.group.startsWith("org.eclipse.jetty") && details.requested.version.startsWith("12")) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -54,6 +54,16 @@ configurations {
       // Force v3.19 of commons-lang3 to avoid CVE-2025-48924 (https://www.cve.org/CVERecord?id=CVE-2025-48924), which
       // is caused by the use of avro-compiler v1.12.0 with older dependencies including commons-lang3 v3.12.0.
       force 'org.apache.commons:commons-lang3:3.19.0'
+
+      // Force v12.1.1 of jetty-http to avoid CVE-2025-5115
+      // (https://nvd.nist.gov/vuln/detail/CVE-2025-5115), which is a transitive
+      // dependency of Kafka connect-runtime:4.1.0
+      force "org.eclipse.jetty:jetty-http:12.1.1"
```
rjrudin, line 61: This is risky because it forces one library to 12.1.1 but doesn't affect other jetty libraries - e.g.

```
+--- org.eclipse.jetty:jetty-server:12.0.22
|    |    +--- org.eclipse.jetty:jetty-http:12.0.22 -> 12.1.1
```

The better approach to make all Jetty stuff use 12.1.1 is this, which I copied from the Flux Gradle file:

```
  configurations.all {
    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
      // Defining this rule for io.netty here so that it affects both flux-cli and flux-embedding-model-azure-open-ai.
      if (details.requested.group.equals("io.netty") and details.requested.version.startsWith("4.1.")) {
        details.useVersion "4.1.127.Final"
        details.because "Bumping from 4.1.122 (what Spark SQL 4.1.0-preview1 depends on) to 4.1.127 to eliminate CVEs."
      }
    }
  }
```

You can say "if group is org.eclipse.jetty and version starts with 12", then we want to use 12.1.1. That will bump all Jetty dependencies. You could possibly do just "if group is org.eclipse.jetty", but I think it's best to include a small check on the version number to make sure you're not modifying anything you don't want to. 

For netty - it looks like just forcing netty-all does the trick. Though I still like the above approach if only because you can use the `because` to document what's going on instead of a comment.

rjrudin (changes requested):

